Issue #75, opened 2014-12-19
Offer a flexible exception policy on request/response validation failures


gchin:
I'm having an issue where raised ResponseValidationError exceptions don't trigger logging in my pyramid app due to the exception not propagating to the tween exception handler because it's an instance of HTTPInternalServerError.  This needs to be an uncaught exception to trigger the default error handler, so simply changing subclass to Exception makes sense to me

PR #80, opened 2015-01-28
Add `validation_context` setting which allows the user to specify a contextmanager to custom handle request/response validation exceptions + docs

gchin:


docs/configuration.rst
```diff
@@ -39,6 +39,11 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: [r'^/static/?', r'^/api-docs/?']
         pyramid_swagger.exclude_paths = [r'^/static/?', r'^/api-docs/?']
 
+        # Definable contextmanager to handle request/response validation
```
striglia: I don't think this will actually work here, since you need to pass a real python object in. Give the example in the python codeblock below?

tests/acceptance/request_test.py
```diff
@@ -154,14 +162,13 @@ def test_200_skip_validation_when_disabled():
         .status_code == 200
 
 
-def test_request_validation_context():
```
striglia: These tests were new in your previous commit, right?

gchin: Yes

striglia:
Fixed #75

PR #84, opened 2015-02-24
Feature: exclude routes from validation

dnephin:
Currently requests can be excluded by path, or by disabling all path validation.

This PR adds support for excluding some routes from validation as another option before disabling all path validation.

y-m uses a single fallback route for legacy routing, so this will allow us to use path validation for all pyramid views, and exclude legacy views.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -39,6 +39,9 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: [r'^/static/?', r'^/api-docs/?']
         pyramid_swagger.exclude_paths = [r'^/static/?', r'^/api-docs/?']
 
+        # Exclude pyramid routes from validation. Accepts a list of strings
+        pyramid_swagger.exclude_routes = ['catchall', 'no-validation']
+
         # Path to contextmanager to handle request/response validation
         # exceptions. This should be a dotted python name as per
         # http://docs.pylonsproject.org/projects/pyramid/en/latest/glossary.html#term-dotted-python-name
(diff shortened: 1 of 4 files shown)


dnephin:
This branch needs the test fix from #82

docs/configuration.rst
```diff
@@ -39,6 +39,9 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: [r'^/static/?', r'^/api-docs/?']
         pyramid_swagger.exclude_paths = [r'^/static/?', r'^/api-docs/?']
 
+        # Exclude routes from validation. Accepts a list of strings
```
striglia: "exclude pyramid routes"? Not sure if this needs clarification...

pyramid_swagger/tween.py
```diff
@@ -97,8 +100,16 @@ def validation_tween_factory(handler, registry):
...
 
     def validator_tween(request):
-        if disable_all_validation or \
-                should_exclude_path(settings.exclude_paths, request.path):
+        # We don't have access to this yet but let's go ahead and build the
+        # matchdict so we can validate it and use it to exclude routes from
+        # validation.
+        route_info = route_mapper(request)
+
+        if (
```
striglia: Could pull out into a baby function. /shrug

striglia:
No major issues. I'll re-review once comments + merging w/ #82 are done

docs/configuration.rst
dnephin: Doesn't hurt to be explicit about it, will update

pyramid_swagger/tween.py
dnephin: ya, I could see that, will do

PR #86, opened 2015-03-03
Delegate spec validation to `swagger_spec_validator`.

### LINKED ISSUE #81: Delegate json schema validation to swagger_spec_validator.
This includes removal of json schema files from the repo, and instead call `swagger_spec_validator.validate()`. This should also resolve #61 and others.

prat0318:
Fixes #81. This commit delegates swagger validation to py package
`swagger_spec_validator`. This commit only covers the validation part of 
swagger 1.2/2.0 spec and in no way handles the serving of 2.0 compatible endpoint.

It currently breaks py33 and py34 tests. This should be rectified with
the merge of Yelp/swagger_spec_validator#11.

Another change made is regarding the swagger spec file convention between 1.2 and 2.0.
Validation check first tries to find Swagger 2.0 spec (which should be just a single file, named, `swagger.json`) otherwise it tries to find Swagger 1.2 spec. `swagger_spec_validator` reads the spec again to decide whether the spec is a 1.2 or a 2.0. I am checking these here because, convention wise, `api_docs.json` is not a correct name for a `2.0` swagger schema. More such changes would need to be made when `pyramid_swagger` handles `2.0` schema completely.


tests/api_test.py
```diff
@@ -24,25 +22,24 @@ def test_basepath_rewriting():
 def build_config(schema_dir):
     return mock.Mock(
         registry=get_registry({
-            'swagger_schema': compile_swagger_schema(schema_dir),
+            'swagger_schema': validate_swagger_schema(schema_dir),
         })
```
dnephin: I don't think this is right `validate_swagger_schema` doesn't return anything. This should still be `compile_swagger_schema`.

`pyramid_swagger` should still return the same exceptions, we probably just want to wrap the exception

dnephin:
This mostly lgtm, but I think the test changes show that we actually need to wrap some exceptions. Users should expect the same exceptions out of `pyramid_swagger`.

tests/api_test.py
prat0318: Oh correct. I was testing my validation spec here which was wrong. I have reverted back these changes, and added similar tests in `spec_test.py` which get validated through `swagger_spec_validator`.

prat0318:
Currently the possible exposed exceptions are `ResourceListingNotFoundError`, `ApiDeclarationNotFoundError` and `SwaggerValidationError`. Either we can use the third one to wrap the first two, or wrap ssv's `SwaggerValidationError` to some `pyramid_swagger`'s `SwaggerValidationError`. I will create a separate issue to track it.

striglia:
Yeah I'd focus on backwards compatibility right now, but keep in mind we're free to have more descriptive exceptions once we do a breaking change. My guess is users won't find distinct exception classes for the various "your spec is invalid" exceptions all that useful....we can probably just use one consistent exception and different messages.

prat0318:
This patch is BC, except that `jsonschema.ValidationError` error thrown is now been replaced with `swagger_spec_validator.SwaggerValidationError`. Do we want to wrap the later error with the former for this patch to maintain BC?

striglia:
Yeah let's wrap it up to maintain backwards compatibility. I don't mind if exception msg changes, but the class should be consistent. You are free to modify the message to make it clear what is going wrong.

prat0318:
Made the change BC by wrapping the ssv's error with `jsonschema`'s `ValidationError`. A subtle difference still remains is that if a `resource_listing` file is not present in the `schema_dir`, earlier `IOError` was thrown, but now `jsonschema`'s `ValidationError` will be thrown (as all ssv's exceptions are wrapped under ValidationError), but i don't think that will be a showstopper.

striglia:
That subtle difference is probably fine. I want to revamp these errors in 2.0.0 but for now we'll let it slide. Looks good aside from the CI build passing.

pyramid_swagger/exceptions.py
```diff
@@ -8,3 +11,12 @@ class RequestValidationError(HTTPClientError):
 
 class ResponseValidationError(HTTPInternalServerError):
     pass
+
+
+def wrap_exception(method):
```
striglia: One thought: let's name this specifically or pass in the exception to reraise as.

analogue:
+1 "we can probably just use one consistent exception and different messages."

pyramid_swagger/spec.py
```diff
@@ -2,54 +2,39 @@
...
+    if os.path.isfile(swagger20_filepath):
+        return swagger20_filepath
+    elif os.path.isfile(swagger12_filepath):
+        return swagger12_filepath
+    else:
+        raise swagger_spec_validator.SwaggerValidationError(
+            'No swagger spec found in directory:{0}. Note that your json file '
+            'must be named {1} or {2}'.format(
+                schema_dir, SWAGGER_2DOT0_FILENAME, API_DOCS_FILENAME))
+
```
analogue: Possible to remove the `else` stmt.

```python
if x:
    return swagger20_filepath
if y:
    return swagger12_filepath
raise swagger_spec_validator.SwaggerValidationError(...)
```

tests/includeme_test.py
```diff
@@ -23,10 +23,20 @@ def test_bad_schema_validated_on_include():
         'pyramid_swagger.schema_directory': 'tests/sample_schemas/bad_app/',
     }
     mock_config = mock.Mock(registry=mock.Mock(settings=settings))
-    with pytest.raises(jsonschema.exceptions.ValidationError):
+    with pytest.raises(ValidationError):
         pyramid_swagger.includeme(mock_config)
```
analogue: I'm not familiar enough with the context here, but I think it is good practice to attempt to assert the message and/or contents of the exception to tighten up the expectation.

```python
with pytest.raises(ValidationError) as excinfo:
    blah
assert 'failed blah blah blah' in str(excinfo.value)
```

tests/sample_schemas/2.0/swagger.json
```diff
@@ -0,0 +1 @@
...nteger","format":"int32"},"type":{"type":"string"},"message":{"type":"string"}},"xml":{"name":"##default"}},"Category":{"properties":{"id":{"type":"integer","format":"int64"},"name":{"type":"string"}},"xml":{"name":"Category"}},"Pet":{"required":["name","photoUrls"],"properties":{"id":{"type":"integer","format":"int64"},"category":{"$ref":"#/definitions/Category"},"name":{"type":"string","example":"doggie"},"photoUrls":{"type":"array","xml":{"name":"photoUrl","wrapped":true},"items":{"type":"string"}},"tags":{"type":"array","xml":{"name":"tag","wrapped":true},"items":{"$ref":"#/definitions/Tag"}},"status":{"type":"string","description":"pet status in the store","enum":["available","pending","sold"]}},"xml":{"name":"Pet"}},"Tag":{"properties":{"id":{"type":"integer","format":"int64"},"name":{"type":"string"}},"xml":{"name":"Tag"}},"Order":{"properties":{"id":{"type":"integer","format":"int64"},"petId":{"type":"integer","format":"int64"},"quantity":{"type":"integer","format":"int32"},"shipDate":{"type":"string","format":"date-time"},"status":{"type":"string","description":"Order Status","enum":["placed","approved","delivered"]},"complete":{"type":"boolean"}},"xml":{"name":"Order"}}}}
```
analogue: Can this not be a single really long line?

pyramid_swagger/spec.py
prat0318: ack.

tests/includeme_test.py
prat0318: ack.

tests/sample_schemas/2.0/swagger.json
prat0318: ack.

pyramid_swagger/exceptions.py
prat0318: Good idea. i will pass the exception to reraise as an arg.

striglia:
@prat0318 is this ready or still in progress?

dnephin:
This is kind of blocked on the `jsonschema` changes as well, to fix the py3 errors.

prat0318:
Yeah, blocked specifically on juian/jsonschema#203 getting merged.

tests/sample_schemas/bad_app/bad_sample.json
```diff
@@ -14,7 +14,8 @@
                         {
                             "paramType": "query",
                             "name": "content",
-                            "required": true
+                            "required": true,
+                            "type": "string"
```
prat0318: type: string was added as jsonschema was flaking with `no type present` error instead of intended `nickname not present` error.

pyramid_swagger/load_schema.py
```diff
@@ -16,6 +16,7 @@
 
 
 EXTENDED_TYPES = {
+    'number': (float,),
```
prat0318: 'float': (float) expected the param type to be 'float', which was not as per swagger spec 1.2 (the format can be float, not the type).

prat0318:
@dnephin @analogue there were few merge conflicts with master, which i resolved. There is a `float` correction and sample tests spec correction as well in this commit. lmk if it looks good to merge.

pyramid_swagger/exceptions.py
```diff
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+import sys
+
+import six
+
```
analogue: no blank line since `six` and `pyramid` are both 3rd party

pyramid_swagger/ingest.py
```diff
@@ -7,10 +7,8 @@
 
 from .load_schema import load_schema
 from .model import SwaggerSchema
-from .spec import validate_swagger_schemas
-
-
-API_DOCS_FILENAME = 'api_docs.json'
+from .spec import (API_DOCS_FILENAME,
+                   validate_swagger_schema)
```
analogue: Looks like this import will fit on a single line

tests/includeme_test.py
```diff
@@ -23,8 +23,19 @@ def test_bad_schema_validated_on_include():
...
+    assert "'nickname' is a required property" in str(excinfo.value)
+
+
+@mock.patch('pyramid_swagger.register_api_doc_endpoints')
+@mock.patch('pyramid_swagger.get_swagger_schema')
+def test_2dot0_schema_validated_on_include(_, _mock):
+    settings = {
+        'pyramid_swagger.schema_directory': 'tests/sample_schemas/2.0/',
+    }
+    mock_config = mock.Mock(registry=mock.Mock(settings=settings))
```
analogue: Please include `spec=<type>` as an argument so it is easy to infer the expected type. This also has the benefit of making sure you don't deviate from the type's expected interface.

analogue:
Fix-n-ship

pyramid_swagger/exceptions.py
prat0318: Ack.

pyramid_swagger/ingest.py
prat0318: Ack.

tests/includeme_test.py
prat0318: Ack.

dnephin:
Looks good!

PR #87, opened 2015-03-04
Optionally generate a resource listing

### LINKED ISSUE #83: Feature: Generate the resource listing
`pyramid_swagger.ingest.build_schema_mapping()` expects a file in a very specific format where the paths to api docs match the filename containing the docs (with the `.json` suffix removed). The rest of the listing file is pretty minimal (a `swaggerVersion` and an `apiVersion`). 

In another project I took the approach of generating this file instead. From the list of files in the directory (plus the two versions) we can create the entire resource listing. 

I propose we do the same in `pyramid_swagger` using `pyramid_swagger.generate_resource_listing` to enable the feature (default it to false for backwards compatibility).  The `apiVersion` and `swaggerVersion` could be specified in a file in the directory.

dnephin:
Resolves #83

Add a new setting `pyramid_swagger.generate_resource_listing` defaulted to False. When enabled the `api_docs.json` would be only a template (containing versions and possibly auth), but no api listings.  The api listing would be generated from the `.json` files in `schema_dir`.

`includeme` was starting to get pretty big after I moved some things into it, so I split it back out into a function in `ingest`.

`model.SwaggerSchema.resource_listing` is not the actual dict of the resource listing, instead of a filepath.


docs/changelog.rst
```diff
@@ -2,7 +2,7 @@ Changelog
 =========
 
 1.5.0-dev (2015-02-25)
-++++++++++++++++++
+++++++++++++++++++++++
 **Under development**
 
 * Added ``enable_api_docs_views`` configuration option so /api-docs
```
striglia: Mention the feature in the changelog?

pyramid_swagger/ingest.py
```diff
@@ -62,27 +82,76 @@ def _load_resource_listing(resource_listing):
...
+    :param schema_dir: the directory which contains swagger spec files
+    :type  schema_dir: string
+    :param should_generate_resource_listing: when True a resource listing will
+        be generated from the list of *.json files in the schema_dir. Otherwise
+        return the contents of the resource listing file
+    :type should_enerate_resource_listing: boolean
+    :returns: the contents of a resource listing document
+    """
+    listing_filename = os.path.join(schema_dir, API_DOCS_FILENAME)
+    resource_listing = _load_resource_listing(listing_filename)
```
striglia: Won't this raise an API error if you don't include one using the new auto-gen feature? Where are you catching that?

docs/changelog.rst
dnephin: Oops, right, will do

pyramid_swagger/ingest.py
dnephin: Even with the auto-gen feature, you still need a "base" resource listing file, that includes all the other keys (everything except for `apis`. Which is at least the `swaggerVersion`, but could also include a few other keys (apiVersion, authentication, etc).

I wasn't sure how to properly express that. Maybe this needs another section in the docs to fully explain how it works (with some examples) ?

Maybe this should be renamed to `generate_api_listing` instead of `generate_resource_listing` ?

striglia: Oh tricky. Yeah I'd express that more clearly in the configuration docs. I'm down for subsections devoted to explaining particularly non-intuitive or complex options. Just add a new one for your new feature and compare the before/after api_docs.json?

As for naming, I don't mind the current one. Maybe ``complete_resource_listing`` if we want to be explicit about the fact that we aren't generating from scratch?

pyramid_swagger/ingest.py
```diff
@@ -62,27 +82,76 @@ def _load_resource_listing(resource_listing):
         )
 
 
-def compile_swagger_schema(schema_dir):
+def generate_resource_listing(schema_dir, listing_base):
+    if 'apis' in listing_base:
```
prat0318: Instead of checking the presence of key `apis`, i would check if it is an empty list. As `apis` is a required field, empty `apis` still makes a valid spec and kinda shows that it is empty for a purpose.

dnephin: I would not want this to be a valid spec, I think that would be more confusing.  This should  be just a template of  the resource listing, without any `apis`.

dnephin:
Pushed full docs + changelog

docs/configuration.rst
```diff
@@ -63,3 +68,67 @@ Note that, equivalently, you can add these during webapp configuration:
...
+
+When ``generate_resource_listing`` is enabled, the ``api_docs.json`` should
+be similar, but with the ``apis`` section removed.
+
+.. code-block:: json
+
+    {
+        "swaggerVersion": "1.2",
+        "apiVersion": "1.0",
+    }
```
striglia: And just to be explicit, do you want to specify what this will generate? Or leave it a little vague so you aren't bound to a particular implementation? Worth calling out that this will not generate any description strings? https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#512-resource-object

dnephin: Good call, I'll add a line to say, "this will generate the equivalent of the original api_docs.json", and add a note about no descriptions

dnephin:
Ok, note added

striglia:
Awesome. Thanks for all the work :)

Issue #22, opened 2014-06-06
Consider having pyramid_swagger handle basic primitive conversion in query/path/bodies


striglia:
One major source of repetitive busywork in Pyramid views is casting query and path parameters into their proper type in your view (e.g. /user/1 will appear in Pyramid's matchdict as a string, not an integer). A service's Swagger spec has the necessary context to do this at the tween level, and in fact we already have the mapping available for doing our own validation. For query bodies, we should just be strongly recommending JSON everywhere as it will handle its own typecasting.

This would simplify boilerplate in views which use these parameters, which often looks like:

``` python
@view_config('user_search')  # Serves /user/{id}/items?count=5&start=0
def user_search_view(request):
    id = int(request.matchdict['id'])
    start = int(request.params['start'])
    count = int(request.params['count'])
    user_item_search(id, count, start)
```

Would offering casting for incoming query and path parameters be of interest? The proposal would be to only support the types defined by Swagger, with no plans to implement richer types (that is up to service authors to do separately). As always, we can make the feature controlled by a setting in your Pyramid app's config.

dnephin:
Looking at the code, I think the casting to type is already being done (for the most part, maybe not date/date-time), it just needs to be added to the request somewhere.  With gluapy I added a dict at `request.data` where the keys were the parameter names, and the values were the already casted python values.  I liked that setup a lot and would be interested in doing the same here.

`request.data` could be some other property, maybe `request.swagger_data` ? Parameter names are unique so there should be no chance of conflicts in the dict.

I like this setup because you still get access to the raw request values in their usual place, and you don't have to deal with extracting values from different parts of the request object. They're all available in the same dict.

striglia:
Good call, https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#524-parameter-object definitely requires the parameter names to be unique.

Yeah this honestly feels like a relatively easy win, particularly if we make it opt-in by not mutating the core request data like you propose. It can pull things off the wire and instantiate basic python types, then library users are free to build different DTOs, use the data directly, whatever.

sabrina261087:
@view_config(route_name="findPetsByStatus")
def findPetsByStatus(request):
    # Returns thing_id as an int (assuming the swagger type is integer)
    thing_id = request.swagger_data['petId']
    return {thing_id}

I tested this method http://localhost:6543/pet/findByStatus?status=available/ but I get 
[...]

PR #88, opened 2015-03-08
Expose Validators instead of schemas and resolvers

### LINKED ISSUE #1: Headers are not validated
Swagger supports header validation (as another paramType) but pyramid_swagger's validators currently ignore them entirely. Fix this!

dnephin:
Resolves #1 - header validation

As I was working on #22 I noticed that I was having to frequently pass around a `resolver`, but that resolver was only ever going to be used with the `schema` that came from the same place.

This refactor exposes a `SchemaValidator` object instead of the the `SchemaAndResolver` object. The validator can be used directly to validate a field from the request or response, but it also stores the original schema which can be used to cast values.

Exposing a validator instead of the schema and resolver should allow us to make changes to the validator (as suggested by some TODOs added in this PR) in a single location (`load_schema`). Previously it would require changes to both `load_schema` and `tween`. 

I think it also makes the tween easier to reason about.  All of the validation logic is now part of the model/load_schema.  The tween only needs to iterate over the fields and validate each without worrying about the implementation of that validation.

### DIFF
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -55,7 +55,8 @@ def build_param_schema(schema, param_type):
         return {
             'type': 'object',
             'properties': properties,
-            'additionalProperties': False,
+            # Allow extra headers
+            'additionalProperties': param_type == 'header',
         }
     else:
         return None
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -10,7 +10,7 @@ import simplejson
 from jsonschema import RefResolver
 
 
-def extract_query_param_schema(schema):
+def build_param_schema(schema, param_type):
     """Turn a swagger endpoint schema into an equivalent one to validate our
     request.
 
@@ -38,7 +38,7 @@ def extract_query_param_schema(schema):
     properties = dict(
         (s['name'], strip_swagger_markup(s))
         for s in schema['parameters']
-        if s['paramType'] == 'query'
+        if s['paramType'] == param_type
     )
     # Generate a jsonschema that describes the set of all query parameters. We
     # can then validate this against dict(request.params).
@@ -52,39 +52,7 @@ def extract_query_param_schema(schema):
         return None
 
 
-def extract_path_schema(schema):
-    """Extract a schema for path variables for an endpoint.
-
-    As an example, this would take this swagger schema:
-        {
-            "paramType": "path",
-            "type": "string",
-            "enum": ["foo", "bar"],
-            "required": true
-        }
-    To this jsonschema:
-        {
-            "type": "string",
-            "enum": ["foo", "bar"],
-        }
-    Which we can then validate against a JSON object we construct from the
-    pyramid request.
-    """
-    properties = dict(
-        (s['name'], strip_swagger_markup(s))
-        for s in schema['parameters']
-        if s['paramType'] == 'path'
-    )
-    if properties:
-        return {
-            'type': 'object',
-            'properties': properties,
-            'additionalProperties': False,
-        }
-    else:
-        return None
-
-
+# TODO: do this with jsonschema directly
 def extract_body_schema(schema, models_schema):
     """Turn a swagger endpoint schema into an equivalent one to validate our
     request.
@@ -128,6 +96,7 @@ def extract_body_schema(schema, models_schema):
         return None
 
 
+# TODO: do this with jsonschema directly
 def strip_swagger_markup(schema):
     """Turn a swagger URL parameter schema into a raw jsonschema.
 
@@ -146,7 +115,7 @@ def strip_swagger_markup(schema):
 
 def get_model_resolver(schema):
     """
-    Gets the schema and a RefResolver. RefResolver's will resolve "$ref:
+    Get a RefResolver. RefResolver's will resolve "$ref:
     ObjectType" entries in the schema, which are used to describe more complex
     objects.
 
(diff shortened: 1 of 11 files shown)


dnephin:
I still need to add an acceptance test for the header validation

pyramid_swagger/tween.py
```diff
@@ -202,50 +185,18 @@ def should_exclude_route(excluded_routes, route_info):
...
-    :param route_match: dict
-    :param request: the request object to validate
-    :type request: Pyramid request object passed into a view
-    :param schema_data: our mapping from request data to schemas (see
-        load_schema)
-    :type schema_data: dict
-    :param resolver: the request object to validate
-    :type resolver: Pyramid request object passed into a view
-    """
+def _validate_request(route_match, request, validator_map):
```
dnephin: I dropped these docstring for now, but this and `_validate_response` is going to be replaced by a decorator in the next branch (for #22).

pyramid_swagger/load_schema.py
```diff
@@ -46,45 +55,14 @@ def extract_query_param_schema(schema):
...
-        for s in schema['parameters']
-        if s['paramType'] == 'path'
-    )
-    if properties:
-        return {
-            'type': 'object',
-            'properties': properties,
-            'additionalProperties': False,
+            # Allow extra headers
+            'additionalProperties': param_type == 'header',
```
striglia: Reasoning?

pyramid_swagger/load_schema.py
```diff
@@ -46,45 +55,14 @@ def extract_query_param_schema(schema):
...
         }
     else:
         return None
 
 
+# TODO: do this with jsonschema directly
```
striglia: I don't think I immediately know what you mean by this. More context for our future selves?

pyramid_swagger/load_schema.py
```diff
@@ -160,49 +139,85 @@ def get_model_resolver(schema):
...
+        ]:
+            args.append(SchemaValidator.from_schema(
+                schema,
+                resolver,
+                validator))
+
+        return cls(*args)
+
+
+class SchemaValidator(object):
```
striglia: Would be good to (somewhere in this class/init/wherever) be specific about the types being passed in.

pyramid_swagger/load_schema.py
```diff
@@ -160,49 +139,85 @@ def get_model_resolver(schema):
...
+        self.schema = schema
+        self.validator = validator
+
+    @classmethod
+    def from_schema(cls, schema, resolver, validator_class):
+        return cls(
+            schema,
+            validator_class(schema, resolver=resolver, types=EXTENDED_TYPES))
+
+    def validate(self, values):
```
striglia: Under what conditions is a schema-less instance of this class valid? Can we just assert in the __init__ and remove this no-op?

pyramid_swagger/load_schema.py
```diff
@@ -160,49 +139,85 @@ def get_model_resolver(schema):
...
+            RequestMatcher(api['path'], operation['method']),
+            ValidatorMap.from_operation(operation, schema_models, resolver)
+        )
+        for api in schema['apis']
+        for operation in api['operations']
+    )
+
+
+class RequestMatcher(object):
+    """Match a :class:`pyramid.request.Request` to a swagger Operation"""
```
striglia: Let's encode (document? assert?) assumptions about the request already being routed (probably better than doing that routing here)

pyramid_swagger/tween.py
striglia: Sounds cool...how so? Decorator on what?

striglia:
Thumbs up on the idea. Looks like a solid refactor whose time has come. Some various clarity/documentation comments but happy to see it wasn't terribly invasive.

Food for thought: is this something we include in the changelog or is it internal-only? I think the latter, but wanted to ask.

pyramid_swagger/tween.py
dnephin: https://github.com/striglia/pyramid_swagger/pull/90/files#diff-03bf6185e91ce5bb05dba792eb334610R260

`validation_error()` decorator around `validate_request()` and `validate_response()`

pyramid_swagger/load_schema.py
dnephin: Ah, because there are often headers outside of the spec for "normal http" requests. Host, User-Agent, Accepts, etc. 

I'll add more detail to this comment.

pyramid_swagger/load_schema.py
dnephin: This is a short-lived TODO, I flushed it out here: https://github.com/striglia/pyramid_swagger/pull/90/files#diff-154107719748d5873abea20526a5c82dR85

pyramid_swagger/load_schema.py
dnephin: True, this needs a better docstring, will do

pyramid_swagger/load_schema.py
dnephin: This is the case when an operation doesn't define any params for the paramType.  We end up with a None schema from `build_param_schema`, so this just skips the validation since we don't have anything to validate.

pyramid_swagger/load_schema.py
dnephin: hmm, does this make that assumption?

I think it just reads the `path` from the request, so it doesn't need to be routed yet.

dnephin:
I  definitely deserves a changelog entry for "adding header validation", but I think the rest is internal-only.

pyramid_swagger/load_schema.py
striglia: Damn good call, my confusion. Drop this.

pyramid_swagger/load_schema.py
striglia: Hmm I see. I suppose we could build an EmptySchema object, but this sounds fine for now.

pyramid_swagger/load_schema.py
striglia: sg2m

dnephin:
Docs are in

dnephin:
Travis failure is the sphinx template name change which is fixed in the other branches

PR #89, opened 2015-03-14
Allow for empty body without headers for void responses

### LINKED ISSUE #79: Void response type requires a body
The `void` response `type` forces a body to be returned, which goes against the spec of HTTP 204 No Content (and other codes that require an empty body):

> The 204 response MUST NOT include a message-body
> --- http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

The problem lies in the `body is None` check:

```python
def validate_outgoing_response(response, schema_map, resolver):
...
    body = prepare_body(response)
    # Short circuit if we are supposed to not validate anything.
    if schema_map.response_body_schema.get('type') == 'void' and body is None:
        return
```

which is can only occur if the body exists and is a JSON encoded `null`:

```python
def prepare_body(response):
    # content_type and charset must both be set to access response.text
    if response.content_type is None or response.charset is None:
        raise HTTPInternalServerError(
            'Response validation error: Content-Type and charset must be set'
        )

    if 'application/json' in response.content_type:
        return simplejson.loads(response.text)
    else:
        return response.text
```

In addition to allowing an empty-body, `void` should also allow the `content_type` and `charset` not to be specified.

dnephin:
Resolves #79 

Addresses two issues:
* a `Context-Type` was expected even for void responses
* a response of an empty dict would fail validation (because it wasn't None)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -93,7 +93,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  Major themes that come with
 # Sphinx are currently 'default' and 'sphinxdoc'.
-html_theme = 'default'
+html_theme = 'classic'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,10 @@ Changelog
   routes which will be ignored for validation purposes.
 * Added ``generate_resource_listing`` configuration option to allow
   pyramid_swagger to generate the ``apis`` section of the resource listing.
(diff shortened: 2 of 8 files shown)


docs/conf.py
```diff
@@ -93,7 +93,7 @@
 
 # The theme to use for HTML and HTML Help pages.  Major themes that come with
 # Sphinx are currently 'default' and 'sphinxdoc'.
-html_theme = 'default'
+html_theme = 'classic'
```
dnephin: Apparently sphinx has changed the name of their default. I had to make this change to get the docs build to pass.

striglia:
Mind adding a bugfix note in the changelog too?

dnephin:
Added a note to the changelog, and also fixed the url warnings for urls that were being redirected now that the github repo has moved.

PR #90, opened 2015-03-14
Primitive conversions

### LINKED ISSUE #22: Consider having pyramid_swagger handle basic primitive conversion in query/path/bodies

dnephin:
Resolves #22 
Branched from #88 so it has a few extra commits.

TODO: missing documentation

### DIFF
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -142,6 +142,14 @@ def get_model_resolver(schema):
     return RefResolver('', '', models)
 
 
+class BodyValidator(Draft4Validator):
+
+    def __init__(self, schema, *args, **kwargs):
+        # TODO: remove this hack once we don't strip name from schema
+        _, schema = schema or (None, None)
+        super(BodyValidator, self).__init__(schema, *args, **kwargs)
+
+
 class ValidatorMap(namedtuple('_VMap', 'query path headers body response')):
     """
     A data object with validators for each part of the request and response
@@ -156,7 +164,7 @@ class ValidatorMap(namedtuple('_VMap', 'query path headers body response')):
             (build_param_schema(operation, 'query'), Draft3Validator),
             (build_param_schema(operation, 'path'), Draft3Validator),
             (build_param_schema(operation, 'header'), Draft3Validator),
-            (extract_body_schema(operation, models), Draft4Validator),
+            (extract_body_schema(operation, models), BodyValidator),
             (extract_response_body_schema(operation, models),
                 Draft4Validator),
         ]:
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -72,7 +72,10 @@ def extract_body_schema(schema):
     return matching_body_schemas[0] if matching_body_schemas else None
 
 
-def noop_validator(_validator, *args):
+def ignore(_validator, *args):
+    """A validator which performs no validation. Used to `ignore` some schema
+    fields during validation.
+    """
     return
 
 
@@ -96,11 +99,19 @@ def required_validator(validator, req, instance, schema):
 
 
 def get_body_validator(models):
+    """Returns a validator for the request body, based on a
+    :class:`jsonschema.validators.Draft4Validator`, with extra validations
+    added for swaggers extensions to jsonschema.
+
+    :param models: a mapping of reference to models
+    :returns: a :class:`jsonschema.validators.Validator` which can validate
+        the request body.
+    """
     return validators.extend(
         Draft4Validator,
         {
-            'paramType': noop_validator,
-            'name': noop_validator,
+            'paramType': ignore,
+            'name': ignore,
             'type': build_swagger_type_validator(models),
             'required': required_validator,
         }
@@ -110,8 +121,8 @@ def get_body_validator(models):
 Swagger12ParamValidator = validators.extend(
     Draft3Validator,
     {
-        'paramType': noop_validator,
-        'name': noop_validator,
+        'paramType': ignore,
+        'name': ignore,
     }
 )
 
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -94,17 +94,19 @@ def extract_body_schema(schema, models_schema):
         for s in schema['parameters']
         if s['paramType'] == 'body'
     ]
-    if matching_body_schemas:
-        schema = matching_body_schemas[0]
-        type_ref = extract_validatable_type(schema['type'], models_schema)
-        # Unpleasant, but we are forced to replace 'type' defns with proper
-        # jsonschema refs.
-        if '$ref' in type_ref:
-            del schema['type']
-            schema.update(type_ref)
-        return strip_swagger_markup(schema)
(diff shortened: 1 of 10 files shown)


pyramid_swagger/load_schema.py
```diff
@@ -35,190 +46,167 @@ def extract_query_param_schema(schema):
...
-            'response_body_schema'
-        ])):
+Swagger12ParamValidator = validators.extend(
+    Draft3Validator,
+    {
+        'paramType': noop_validator,
+        'name': noop_validator,
+    }
+)
+
```
dnephin: This "use custom jsonschema.Validator for validation" stuff isn't really essential.  So if you think it's less-intuitive than the `strip_swagger_markup()`, I have no problem going back to that.

I do think it's kind of cool that we can create swagger validators out of jsonschema pretty easily. I'm not sure what the difference was between Draft3 and Draft4, but it's possible we could change this to Draft4 with other modifications.

tox.ini
```diff
@@ -19,6 +19,7 @@ commands =
 
 [flake8]
 exclude = .tox,*.egg
+max-complexity = 9
```
dnephin: Most of the codebase is actually only complexity 5, it's only `validation_tween_factory` that requires a 9 I think.

pyramid_swagger/load_schema.py
striglia: I think it's simpler and less likely to be missing minor details. Also denser though, so I'll see if I can point out the places where I'm not following and see if we can make it clearer...

pyramid_swagger/load_schema.py
```diff
@@ -44,100 +46,74 @@ def build_param_schema(schema, param_type):
...
+            'required': required_validator,
+        }
     )
-    return RefResolver('', '', models)
+
+
+Swagger12ParamValidator = validators.extend(
+    Draft3Validator,
+    {
+        'paramType': noop_validator,
```
striglia: Validating these fields wouldn't be hard. Maybe we should? Or is this pointless because they should be validated by our spec validation and individual requests ignore them?

pyramid_swagger/tween.py
```diff
@@ -103,21 +108,89 @@ def validator_tween(request):
...
+    @property
+    def headers(self):
+        return self.request.headers
+
+    @property
+    def body(self):
+        return getattr(self.request, 'json_body', {})
+
+
+def handle_request(request, validation_context, validator_map):
```
striglia: Very clean and readable. Damn fine job :)

pyramid_swagger/tween.py
```diff
@@ -103,21 +108,89 @@ def validator_tween(request):
...
+    for validator, values in [
+        (validator_map.query, request.query),
+        (validator_map.path, request.path),
+        (validator_map.headers, request.headers),
+    ]:
+        values = cast_params(validator.schema, values)
+        validation_pairs.append((validator, values))
+        request_data.update(values)
+
+    if validator_map.body.schema:
```
striglia: A little weird that we special case this since it's not the only type of schema that might be empty...

pyramid_swagger/load_schema.py
dnephin: I believe they should already be validated by swagger_spec_validator, but now I'm not as sure. Need to think about this some more.

pyramid_swagger/tween.py
dnephin: I agree. I really tried to get this inline, but in the end it's special-cased because the name is external to the body itself, where as the other `values` contain the key in the dict.

I should add a comment about that

striglia: Oh but you're okay there because `name` must be "body" for the body paramType. https://github.com/swagger-api/swagger-spec/blob/482fb30fd31bfa34bade410f57c48aa3ab268b2b/versions/1.2.md#524-parameter-object

If that's the only reason, I'd say let's just depend on the spec and treat it identically. I guess it's a little odd to write the code...cast_params has to know how to infer it is looking at a body?

pyramid_swagger/load_schema.py
striglia: I'm pretty comfortable leaving it as-is for now. None of the noops I see are request-based, and we can always switch this up later.

striglia:
Let's hash out what we want to do about the body validation special case and otherwise this looks great :) Think it's solid enough to document and changelog? I've been tossing around the idea of tagging a RC for 1.5 as we're including several significant features. If we did that, I'd be willing to tag it as soon as we get through this round of major features.

pyramid_swagger/load_schema.py
dnephin: I thought about it some more, and I think what this is really saying is `ignore` instead of `noop`.  Basically we have these fields in the swagger schema, and they don't actually represent anything about validation, they're for other parts of swagger, so for the purposes of validation we're ignoring them.

I'm going to change the name of this validator to `ignore` instead of `noop_validator`.

pyramid_swagger/tween.py
dnephin: Ya, I can rework the values to make them fit validation, but it breaks on `cast_params`.

dnephin:
Added some more docstring and comment, and added a section to quickstart.rst.

pyramid_swagger/tween.py
striglia: Cool, let's not get hung up on it. Internal detail we can alter later.

pyramid_swagger/load_schema.py
```diff
@@ -44,100 +46,85 @@ def build_param_schema(schema, param_type):
...
 
-def get_model_resolver(schema):
+def required_validator(validator, req, instance, schema):
+    """Swagger 1.2 expects `required` to be a bool in the Parameter object, but
+    a list of properties in a Model object.
     """
-    Get a RefResolver. RefResolver's will resolve "$ref:
-    ObjectType" entries in the schema, which are used to describe more complex
-    objects.
+    if schema.get('paramType') and req is True:
```
prat0318: If `schema.get('paramType')` is `True` but `req` is `False`, in that case it should not call `required_draft4` with `req` with a boolean value, as the method expects a list.

dnephin: You're right, this is a bug.

PR #93, opened 2015-03-27
Add support for validating form-encoded bodies

### LINKED ISSUE #53: Support form-encoded bodies in requests and responses
Currently the code assumes either string or json encodings. It may work currently, but tests should be added to prove it works on both incoming and outgoing content.

johnschultz:
Adds support for validating arguments of `"paramType": "form"`. Fixes #53. 

Looking for some feedback to make sure I'm doing it right and not missing anything.

### DIFF
--- a/tests/acceptance/request_test.py
+++ b/tests/acceptance/request_test.py
@@ -30,6 +30,29 @@ def validation_context(request, response=None):
 validation_ctx_path = 'tests.acceptance.request_test.validation_context'
 
 
+def test_200_with_form_params(test_app):
+    assert test_app.post(
+        '/post_with_form_params',
+        {'form_param': 42},
+    ).status_code == 200
+
+
+def test_400_with_form_params_wrong_type(test_app):
+    assert test_app.post(
+        '/post_with_form_params',
+        {'form_param': "not a number"},
+        expect_errors=True,
+    ).status_code == 400
+
+
+def test_400_if_json_body_for_form_parms(test_app):
+    assert test_app.post_json(
+        '/post_with_form_params',
+        {'form_param': 42},
+        expect_errors=True,
+    ).status_code == 400
+
+
 def test_400_if_required_query_args_absent(test_app):
     assert test_app.get(
         '/sample/path_arg1/resource',
@@ -105,6 +128,7 @@ def test_200_on_json_body_without_contenttype_header(test_app):
     assert test_app.post(
         '/sample?optional_string=bar',
         simplejson.dumps({'foo': 'test'}),
+        {'Content-Type': ''},
     ).status_code == 200
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ build/
 dist/
 .coverage
 .tox
+.ropeproject
 pyramid_swagger.egg-info
 tests/__pycache__
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -127,7 +127,9 @@ Swagger12ParamValidator = validators.extend(
 )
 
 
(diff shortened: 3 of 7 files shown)


tests/acceptance/request_test.py
```diff
@@ -30,6 +30,21 @@ def validation_context(request, response=None):
...
+
+
+def test_400_with_form_params_wrong_type(test_app):
+    assert test_app.post(
+        '/post_with_form_params',
+        {'form_param': "not a number"},
+        expect_errors=True,
+    ).status_code == 400
+
+
```
johnschultz: Do we need more tests than this?

tests/acceptance/request_test.py
```diff
@@ -99,6 +114,7 @@ def test_400_if_required_body_is_missing(test_app):
     ).status_code == 400
 
 
+@pytest.mark.xfail
```
johnschultz: I'm not sure what the right way to fix this test is. It looks like calling `request.POST` before `request.json` causes the second property to fail because the body can't be re-interpreted as json after it's read as form encoded. However, accessing the attributes in the opposite order succeeds. Thoughts?

dnephin:
This looks right to me

tests/acceptance/request_test.py
dnephin: Ah ya, this is an annoying part of the pyramid request object. If we can trust the content-type header, you could add a check for that in the PyramidRequest object, and not actually access the `POST` if the context is json.

striglia: Any reason we cannot trust it? Is a http request with conflicting content
type something we realistically need to protect against?

We can also not use the .json convenience method...

dnephin: Not that it would conflict, but it could be absent possibly (but maybe not?)

johnschultz: Another option is to move the code that validates `json_body` before the code that validates the other parameters. This seems bad because of the  dependency on checking the `json_body` before `body`, but on the other hand, it seems like reading one is mutually exclusive with reading the other. 

I'm hesitant to say we can trust the content-type header. Isn't the whole point of this test as it is originally written that we can not trust it?

dnephin: We could just treat "no content-type" as json, and only read `request.POST` if the header is one of 'application/x-www-form-urlencoded', 'multipart/form-data'

I believe a proper content type is going to be required for that anyway, right?

So form could be 

```python

@property 
def form(self):
    if self.request.headers.get('Content-Type') in FORM_TYPES:
       return request.POST
    return {}
```

striglia: Works for me

johnschultz: Ok, I'll add. Do we want to be very strict, and also require the content-type value for json as well?

dnephin: I think we'll want to preserve the existing behaviour of this test, so defaulting to `json` when it's missing.

johnschultz:
I think everything is good now if someone wants to merge and release.

tests/acceptance/request_test.py
striglia: Confirm that a wrongly-typed body (e.g. JSON) gets rejected?

striglia:
Sadly this just missed my 1.5.0-rc1 release, but we can put out an rc2 easily enough, as this has been an often requested feature.

tests/acceptance/request_test.py
johnschultz: Added.

johnschultz:
Updated with another test.

striglia:
Merged! Thanks :)

Issue #94, opened 2015-03-29
Epic: Swagger 2.0 support


dnephin:
I thought I'd create this ticket as a tracking ticket for all the swagger 2.0 work.

**Important**:
All version compatibility logic (`if v1.2 do x, else do y`) should  be kept to a single module instead of cluttering the whole project with these switches.  Right now `pyramid_swagger` is well setup for this. All the logic switches can be kept to the `pyramid_swagger.ingest` module.

**Tasks**:
- [x] `pyramid_swagger.ingest.get_swagger_schema(...)` needs to be able to examine the schemas in the `schema_directory`, determine which swagger version it is, and return a `pyramid_swagger.model.SwaggerSchema` loaded with the swagger 2.0 schemas, but which conforms to the same interface as the swagger 1.2 schema.  (Interfaces changes for `api_declarations` addressed in #95)
- [x] api docs validation, should be handled by #81
- [x] `pyramid_swagger.tween.validation_tween_factory` should be largely unchanged, the validators it uses should support the same interface. Required changes are:
  * `cast_params()`  reads the schemas, so it may need an abstraction to work with both versions.
  * `validate_response()` needs to change because `void` doesn't appear to be a type any more
  * `handle_request` reads the body param name from the schema, but it could just be a constant
- [x] `pyramid_swagger.api.register_api_doc_endpoints()` will need to be changed to just return any file in the directory for 2.0. (based on #95)

I think the bulk of the work is really the first task. Once we have an interface that hides the differences between the specs it should be easy to integrate without adding any branches to handled the differences between spec versions.

striglia:
Sounds pretty solid to me. As you say, the main work is transparently discovering and ingesting the 1.2 or 2.0 schema without explicit specifying. Bravado-core should solve both spec validation and at least providing the interface for request/response validation. At that point, for the purposes of validation pyramid_swagger just has to interface with bravado-core.

Ideally I'd like `register_api_doc_endpoints` to just accept the list of ingested schemas (which will contain their versions) rather than the current model of re-ingesting them. The re-ingest model seems a little dated at this point.

analogue:
After having a good look at pyramid-swagger, I don't think this approach is viable.  I'll go through my initial list of road blocks:

1) "All the logic switches can be kept to the pyramid_swagger.ingest module."
Ingestion is not the only place 1.2 and 2.0 differ in logic. It is pervasive and far from easy to keep isolated.

[...]

PR #95, opened 2015-03-29
Refactor api_doc registering to support 2.0

dnephin:
This addresses task 4 in #94 

It starts to add support for a things:
* it will allow multiple sets of api_docs to be registered by calling `pyramid_swagger.api.register_api_doc_endpoints`. This will be necessary for our transition to swagger 2.0. The plan is currently to serve api docs for both swagger 1.2 and 2.0, but only validate against the 2.0 specs.  This way both sets of clients will continue to work.
* allows for the base_path of the api docs to be customized, this is kind of necessary if we're registering multiple sets of docs
* provides a method on the `SwaggerSchema` interface that can be used to abstract away the difference between swagger 1.2 and 2.0 api docs. This will allow us to keep all the version switches in a single module (`pyramid_swagger.ingest`).

I've left the documentation and settings changes as TODOs for now so we can discuss this further and make sure this is the right approach.

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -4,26 +4,41 @@ Module for automatically serving /api-docs* via Pyramid.
 """
 import simplejson
 
+from pyramid_swagger.model import PyramidEndpoint
 
-def register_api_doc_endpoints(config):
-    """Create and register pyramid endpoints for /api-docs*."""
-    swagger_schema = config.registry.settings['pyramid_swagger.schema']
-    register_resource_listing(config, swagger_schema.resource_listing)
 
-    for name, filepath in swagger_schema.api_declarations.items():
+# TODO: document that this is now a public interface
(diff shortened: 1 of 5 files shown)


pyramid_swagger/api.py
```diff
@@ -4,26 +4,41 @@
...
-    register_resource_listing(config, swagger_schema.resource_listing)
 
-    for name, filepath in swagger_schema.api_declarations.items():
-        with open(filepath) as input_file:
-            register_api_declaration(
-                config,
-                name,
-                simplejson.load(input_file)
-            )
+# TODO: document that this is now a public interface
```
striglia: "this" is the fact that we're serving data at /api-docs? Or that it can be customized?

dnephin: `register_api_doc_endpoints` would be a public interface so that users can register multiple sets of api docs at different base_paths.

PR #100, opened 2015-04-08
Rough first draft for Swagger 2.0 request support

analogue:
This is a rough first cut for routing request handling through a tween to bravado-core. 

Notes:
- Swagger 1.2 is not in the code path at all.
- `tween20.py` created to isolate Swagger 2.0 support for the time being
- Depends on https://github.com/Yelp/swagger-py/pull/140 being merged
- Manually tested using example_happyhour and works wonderfully
- Couldn't get `dependency_links` working in `setup.py` so unit tests still a WIP

Will re-submit when a `swagger2.0` branch is created upstream.


analogue:
I'm going to get the bravado-core package in order so this no longer fails building.

pyramid_swagger/ingest.py
```diff
@@ -154,6 +156,33 @@ def get_swagger_schema(settings):
...
+    the swagger specs in `pyramid_swagger.schema_directory`. If
+    `pyramid_swagger.enable_swagger_spec_validation` is enabled the schema
+    will be validated before returning it.
+
+    :param settings: a pyramid registry settings with configuration for
+        building a swagger schema
+    :type settings: dict
+    :rtype: :class:`bravado_core.spec.Spec`
+    """
+    schema_dir = settings.get('pyramid_swagger.schema_directory', 'api_docs/')
```
prat0318: I think `swagger/` will be a better default location than `api_docs/`

pyramid_swagger/ingest.py
```diff
@@ -154,6 +156,33 @@ def get_swagger_schema(settings):
...
+    with open(os.path.join(schema_dir, 'swagger.json'), 'r') as f:
+        spec_dict = simplejson.loads(f.read())
```
prat0318: Note: I am planning a later PR to check for an existance for a `swagger.json`. If it is not present combine individual swagger specs present in the directory, similar to something like this: http://pastebin.com/TzKAtKEA

pyramid_swagger/__init__.py
```diff
@@ -4,20 +4,28 @@
...
     settings = config.registry.settings
 
     # Add the SwaggerSchema to settings to make it avialable to the validation
     # tween and `register_api_doc_endpoints`
     settings['pyramid_swagger.schema'] = get_swagger_schema(settings)
+    settings['pyramid_swagger.spec'] = get_swagger_spec(settings)
 
     config.add_tween(
-        "pyramid_swagger.tween.validation_tween_factory",
+        #"pyramid_swagger.tween.validation_tween_factory",   # 1.2
```
prat0318: I'm curious if with the above comment, will `pyramid_swagger` be non-compatible to v1.2?

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,216 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from collections import namedtuple
+import logging
+
+from pyramid.interfaces import IRoutesMapper
+
+from bravado_core.request import RequestLike, unmarshal_request
```
prat0318: Nit: above should be combined with the 3rd party package group.

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,216 @@
...
+            return handler(request)
+
+        validation_context = _get_validation_context(registry)
+
+        try:
+            swaggerize_request(request, settings, route_info)
+        except (PathNotMatchedError, RequestValidationError) as exc:
+            if settings.validate_path:
+                with validation_context(request):
+                    raise RequestValidationError(str(exc))
```
prat0318: Does it maintain the original trace?

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,216 @@
...
+
+        if should_exclude_request(settings, request, route_info):
+            return handler(request)
+
+        validation_context = _get_validation_context(registry)
+
+        try:
+            swaggerize_request(request, settings, route_info)
+        except (PathNotMatchedError, RequestValidationError) as exc:
+            if settings.validate_path:
```
prat0318: `RequestValidationError` generated should not be ignored if `validate_path` is set to `False`.

pyramid_swagger/__init__.py
dnephin: I'm pretty sure this branch is just a preview to show  it working with 2.0, it wont be merged this way

pyramid_swagger/api.py
```diff
@@ -74,3 +74,25 @@ def view_for_api_declaration(request):
...
+        # without file IO at request time.
+        return spec.spec_dict
+
+    route_name = 'swagger_json'
+    config.add_route(route_name, '/swagger.json')
+    config.add_view(
+        view_for_swagger_json,
+        route_name=route_name,
+        renderer='json'
+    )
```
dnephin: This is heavily duplicated with above, please use #95 to address this.

pyramid_swagger/tween20.py
analogue: ack

pyramid_swagger/tween20.py
analogue: That would be nice!

pyramid_swagger/api.py
analogue: Registering multiple api-docs endpoints is temporary until 1.2 support is eventually removed. I don't see the benefit of all the changes in #95 when the simple change above suffices.

pyramid_swagger/tween20.py
analogue: Working on tests for all this stuff.

pyramid_swagger/api.py
dnephin: * 1.2 support won't be removed for quite a while, it requires every client to upgrade, that is a slow process
* #95 is not just for 1.2 support, it also makes api registering more flexible, and allows users of the library to do things their own way instead of being forced into only one way. This is important for a successful open source library.

analogue: This pull request is for 2.0 support, not 1.2 and 2.0 co-existing.  If #95 is useful during the 1.2 + 2.0 co-existing effort, then that is wonderful.  I've separated these concerns for many reasons.  It makes things simpler, allows iteration, and produces results.

pyramid_swagger/__init__.py
analogue: It will serve docs for 1.2 and 2.0 but all requests will be routed through the 2.0 implementation.  The catch being that view callables have to work for both versions without any changes.

pyramid_swagger/__init__.py
```diff
@@ -2,22 +2,51 @@
...
+        `jsonref.dumps` when called with a `cls` parameter. Strip the `default`
+        keyword in this special case.
+        """
+        kwargs['default'] = None
+        return jsonref.dumps(obj, **kwargs)
+
+    config.add_renderer(
+        'jsonref',
+        JSON(serializer=dumps_wrapper,
+             cls=jsonref._ref_encoder_factory(json.JSONEncoder)))
```
dnephin: (if this is still WIP feel free to ignore this comment)

This seems like the wrong place to handle this.  We're 2 libraries away from the problem.

1. we could fix jsonref to properly support the interface it should be supporting. It claims "Drop in replacement for json.load()", this is a lie.
2. we could handle it in bravado_core to properly return already expanded json

analogue: This is mostly to get this branch building green.  The issue with other 3rd party libraries not being able to handle json with embedded jsonref proxy objects is an issue in some other places too.  I've already decided to strip the jsonref proxy objects in bravado-core.

analogue:
Re-submitting against the swagger2.0 branch.

PR #101, opened 2015-04-21
Support for Swagger 2.0 requests

analogue:
- Swagger 1.2 is not in the code path at all.
- tween20.py created to isolate Swagger 2.0 support
- Manually tested using example_happyhour and works wonderfully
- using requirements.txt for deps until packaged on pypi


analogue:
@dnephin Think this is ready to go on the request side.  I have another branch which is for the response side of things. Let me know what you think.

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,224 @@
...
+class Settings(namedtuple(
+    'Settings',
+    [
+        'spec',
+        'validate_request',
+        'validate_response',
+        'validate_path',
+        'exclude_paths',
+        'exclude_routes',
+    ]
```
dnephin: This looks identical to  the one in `tween.py`, (except for spec vs schema name). Could we use the one in tween.py for now, and rename it to `schema` to `spec` at some point later?

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,224 @@
...
+
+        return response
+
+    return swagger_tween
+
+
+class PyramidSwaggerRequest(RequestLike):
+    """Adapter for a :class:`pyramid.request.Request` which exposes request
+    data for unmarshaling, transformation, and validation.
+    """
```
dnephin: This class is very similar to `tween.py`. There are some bug fixes in this version (`query`) and there are some things missing (checks for context type in `form`).

If you were to replace `body` in the `tween.py` version with this  `json()` and add

```python
body = property(json)
```

We could have just a single `PyramidSwaggerRequest`

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,224 @@
...
+        validation_context = _get_validation_context(registry)
+
+        try:
+            swaggerize_request(request, settings, route_info)
+        except (PathNotMatchedError, RequestValidationError) as exc:
+
+            # raise existing RVE or allow validation_context to override
+            if isinstance(exc, RequestValidationError):
+                with validation_context(request):
+                    raise
```
dnephin: These seem to be the only three lines that are different from the `validation_tween_factory` in significant way. 

I think this could actually be identical with a few minor changes:

* instead of calling swagger_request, it could call `op = get_op_for_request(request, route_info, settings.spec)` which would only raise the `PathNotMatchedError`
* `swaggerize_request` could be called in place of where `handle_request` is in `validation_tween_factory`

I think with those small changes, you could actually use the exact same tween, the only different would be parametrization of:
* `get_op_for_request` vs `settings.schema.validators_for_request`
* `handle_request` vs `swaggerize_request`
* `swaggerize_response` vs `swaggerize_response`

I think that could be done with a `namedtuple('ResponseHandler', 'op_for_request handle_request handle_response')` added to settings.

That way there could be only one tween.

pyramid_swagger/tween20.py
```diff
@@ -0,0 +1,224 @@
...
+    if hasattr(route, 'path'):
+        op = spec.get_op_for_request(request.method, route.path)
+        if op is not None:
+            return op
+    raise PathNotMatchedError(
+        "Could not find a matching Swagger operation for {0} request {1}"
+        .format(request.method, request.url))
+
+
+def load_settings(registry):
```
dnephin: I think with the proposed changes to `Settings` you could also drop this, since it would be identical.

.travis.yml
```diff
@@ -3,8 +3,9 @@ language: python
 env:
 - TOX_ENV=py26
 - TOX_ENV=py27
-- TOX_ENV=py33
-- TOX_ENV=py34
+# TODO: re-enable when bravado-core supports py3.x
+#- TOX_ENV=py33
+#- TOX_ENV=py34
```
dnephin: Would you mind opening an issue for this on the bravado_core github issue tracker and linking it here?

It would be good to restore this support before merging into master, but if there are serious blockers, then I could see an argument for holding off. At least knowing what the blockers were would be good.

dnephin:
I think this looks good. With the proposed changes it could be just three functions (`get_op_for_request`, `swaggerize_request`, and `swaggerize_response`) with some very minor changes to the existing code to parametrize it.

pyramid_swagger/tween20.py
analogue: ack

pyramid_swagger/tween20.py
analogue: ack

pyramid_swagger/tween20.py
analogue: ack

.travis.yml
analogue: I actually did a timeboxed task of porting bravado_core to py3.x.  Turned out to be a lot more work than I expected so I plan to pick that branch up later. Issues arose mostly because of things that don't exist in py3.x like the `long` type if I remember correctly.

https://github.com/Yelp/bravado-core/compare/master...analogue:py3x_support?expand=1


Ticketed @ https://github.com/Yelp/bravado-core/issues/10

pyramid_swagger/tween.py
```diff
@@ -108,10 +115,11 @@ def validator_tween(request):
...
 
         if settings.validate_request:
-            request_data = handle_request(
+            request_data = swagger_handler.handle_request(
                 PyramidSwaggerRequest(request, route_info),
-                validation_context,
-                validator_map)
+                validation_context=validation_context,
+                validator_map=op_or_validators_map,
+                op=op_or_validators_map)
```
dnephin: This is passing the same argument in twice, why not just use positional args and pass it in once?

pyramid_swagger/tween.py
```diff
@@ -122,14 +129,16 @@ def swagger_data(_):
 
         if settings.validate_response:
             with validation_context(request, response=response):
-                validate_response(response, validator_map.response)
+                swagger_handler.handle_response(
+                    response,
+                    validator=getattr(op_or_validators_map, 'response', None))
```
dnephin: I think we could remove this `getattr` by adjusting the `validate_response()` to take the `op_or_validators_map` instead, but that seems fine to wait until the response branch.

dnephin:
lgtm

PR #109, opened 2015-05-05
Service both Swagger 1.2 and 2.0 clients 

analogue:
Still have some unit tests to write but the core of the functionality is there.

General notes:
- Merged in dnephin's branch from https://github.com/striglia/pyramid_swagger/pull/95
- I used a parameterized pytest fixture in `tests/acceptance/request_test.py` to exercise code paths for the variations of `pyramid_swagger.swagger_versions`. Was not able to do with with `response_test.py` hence `response20_test.py`
- Tested with a pyramid app using swaggerpy as a 1.2 client, and bravado as a 2.0 client

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -37,14 +37,15 @@ def build_swagger_12_endpoints(resource_listing, api_declarations):
     :type api_declarations: dict
     :rtype: iterable of :class:`pyramid_swagger.model.PyramidEndpoint`
     """
-    yield build_resource_listing(resource_listing)
+    yield build_swagger_12_resource_listing(resource_listing)
 
     for name, filepath in api_declarations.items():
         with open(filepath) as input_file:
-            yield build_api_declaration(name, simplejson.load(input_file))
+            yield build_swagger_12_api_declaration(
+                name, simplejson.load(input_file))
 
 
-def build_resource_listing(resource_listing):
+def build_swagger_12_resource_listing(resource_listing):
     """
     :param resource_listing: JSON representing a Swagger 1.2 resource listing
     :type resource_listing: dict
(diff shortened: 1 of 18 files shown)


pyramid_swagger/api.py
```diff
@@ -4,66 +4,85 @@
...
+    :param config: a pyramid configuration to register the new views and routes
+    :type  config: :class:`pyramid.config.Configurator`
+    :param endpoints: a list of endpoints to register as routes and views
+    :type  endpoints: a list of :class:`pyramid_swagger.model.PyramidEndpoint`
+    :param base_path: the base path used to register api doc endpoints.
+        Defaults to `/api-docs`.
+    :type  base_path: string
+    """
+    for endpoint in endpoints:
+        path = base_path + endpoint.path
```
prat0318: It would be a good idea to `basepath.rstrip('/')` when allowing user to enter custom basePath

pyramid_swagger/tween.py
```diff
@@ -298,23 +304,21 @@ def build_swagger_handler(settings, schema):
...
             handle_response=swaggerize_response,
         )
-    elif swagger_version == '1.2':
+
+    if SWAGGER_12 in swagger_versions:
         return SwaggerHandler(
             op_for_request=schema.validators_for_request,
             handle_request=handle_request,
             handle_response=validate_response,
         )
```
prat0318: I think we still need the TypeError when any of the versions is not present.

prat0318:
lgtm. I think we can also add a note somewhere mentioning that if both [1.2, 2.0] is present in the settings, then `pyramid_swagger` does the request/response validation completely as per the 2.0 schema

pyramid_swagger/api.py
analogue: ack

pyramid_swagger/tween.py
analogue: :-)

analogue:
Good point. The users should know that the 1.2 and 2.0 schemas are expected to be in sync.

dnephin:
lgtm

Issue #111, opened 2015-05-12
Figure out why master is failing on Travis


striglia:
Worked locally, and the change is not interesting enough to be responsible. It's passed previously, so I'm suspecting something has changed out from under us. Possibly on the Travis side?

SSV merge worked: https://travis-ci.org/striglia/pyramid_swagger/builds/60266076
1.5.0 on py3x failed: https://travis-ci.org/striglia/pyramid_swagger/builds/62315510
Rebuild failed: https://travis-ci.org/striglia/pyramid_swagger/builds/62315520

Roughly a 15 day gap between the clean and failing builds...

prat0318:
This is weird. Travis shows codec issue at the [line](https://github.com/striglia/pyramid_swagger/blob/master/setup.py#L13) which reads from `README.rst` but it has not been changed in this commit. I would try a build again today as it could be a Travis issue.

striglia:
Just rebuilt the previously-green [SSV merge](https://travis-ci.org/striglia/pyramid_swagger/builds/60266076) and it broke in the same way. I'm calling this as a Travis change in the intervening time. Googled around a bit and couldn't identify what's up. 

Seems suspicious that we're doing a py3x test using a py2x interpreter, but I'm not sure whether that's changed recently.

https://travis-ci.org/striglia/pyramid_swagger/jobs/60266079#L299

prat0318:
It is because Travis is now using `tox` version `2.0.0` instead of earlier `1.9.2`.

On my local:
```
(mypy)dev32-devc➜ pyramid_swagger:(master)U  /nail/home/agarwal/mypy/bin/tox --version            [12:29:15]
2.0.0 imported from /nail/home/agarwal/mypy/lib/python2.6/site-packages/tox/__init__.pyc
(mypy)dev32-devc➜ pyramid_swagger:(master)U  /nail/home/agarwal/mypy/bin/tox -e py33              [12:29:24]
GLOB sdist-make: /nail/home/agarwal/srv/pyramid_swagger/setup.py
py33 create: /nail/home/agarwal/srv/pyramid_swagger/.tox/py33
py33 installdeps: coverage, mock, pytest, webtest
py33 inst: /nail/home/agarwal/srv/pyramid_swagger/.tox/dist/pyramid_swagger-1.5.0.zip
ERROR: invocation failed (exit code 1), logfile: /nail/home/agarwal/srv/pyramid_swagger/.tox/py33/log/py33-2.log
[...]

PR #112, opened 2015-05-13
Pin Tox version to 1.9.2 to prevent Travis picking up 2.0.0

prat0318:
Builds like [this](https://travis-ci.org/striglia/pyramid_swagger/builds/62315510) were failing as Travis was picking 2.0.0 tox version and throwing encoding errors. Still need to figure out the root cause.


striglia:
Resolves #111

Issue #113, opened 2015-05-21
multipart/form-data is broken


analogue:
In ```pyramid_swagger.tween.PyramidSwaggerRequest```, form data that is passed as ```multipart/form-data``` always has a ```None``` value because ```self.request.headers.get('Content-Type')``` is a string like ```'multipart/form-data; boundary=----------------------------3d7ce2e63932'``` and not ```'multipart/form-data'```. 

The ```content-type``` should be accessed directly from ```request.content_type``` instead of poking about in the headers.

Relevant code:
```
    @property
    def form(self):
        """
        :rtype: dict
        """
        # Don't read the POST dict unless the body is form encoded
        if self.request.headers.get('Content-Type') in self.FORM_TYPES:
            return self.request.POST.mixed()
        return {}
```

Issue #99, opened 2015-04-01
"File" paramType causes validation to fail


johnschultz:
Looks like fixing form validation exposed a bug with parameters of type `File`. This is what happens when I run a test that uses webtest's upload_files arg (http://webtest.readthedocs.org/en/latest/testapp.html#making-requests) with a `File` type argument in swagger.
### Swagger:

``` JSON
        {
            "path": "/business/{business_id}/media/photo/v1",
            "operations": [
                {
                    "summary": "Upload a business photo",
                    "notes": "<a href='mock_responses/media_examples/business_{business_id}_media_photo_v1_response.POST.json'>Example response</a>",
                    "method": "POST",
                    "nickname": "createPhoto",
                    "parameters": [
                        {
                            "name": "business_id",
                            "type": "string",
                            "description": "Business identifier",
                            "required": true,
                            "paramType": "path"
                        },
                        {
                            "name": "photo",
                            "type": "File",
                            "description": "Binary photo data",
                            "required": true,
                            "paramType": "form"
                        },
                        {
                            "name": "caption",
                            "type": "string",
                            "description": "Photo caption",
                            "required": false,
                            "paramType": "form"
                        }
                    ],
                    "type": "BusinessPhotoResponse",
                    "consumes": ["multipart/form-data"]
                }
            ]
        },
```
### Failure:

``` python
tests/integration_tests/api/business/media_integration_test.py ...Traceback (most recent call last):
  File "/nail/home/jschultz/pg/biz_app/.tox/py/local/lib/python2.7/site-packages/pyramid/tweens.py", line 21, in excview_tween
    response = handler(request)
  File "/nail/home/jschultz/pg/biz_app/.tox/py/local/lib/python2.7/site-packages/pyramid_swagger/tween.py", line 114, in validator_tween
    validator_map)
  File "/nail/home/jschultz/pg/biz_app/.tox/py/local/lib/python2.7/site-packages/pyramid_swagger/tween.py", line 204, in handle_request
    validate_request(validation_pairs)
  File "/nail/home/jschultz/pg/biz_app/.tox/py/local/lib/python2.7/site-packages/pyramid_swagger/tween.py", line 286, in _validate
    raise exc_class(str(exc))
RequestValidationError: 'photo' is a required property

Failed validating 'required' in schema[u'properties']['photo']:
    {u'additionalProperties': False,
     u'properties': {'caption': {'description': 'Photo caption',
                                 'name': 'caption',
[...]

PR #116, opened 2015-05-21
Fix multipart/form-data aka file uploads

analogue:
Fix for #113 and probably #99 too.

Will update changelog, version, etc when PR gets the thumbs up.


pyramid_swagger/load_schema.py
```diff
@@ -88,6 +88,15 @@ def swagger_type_validator(validator, ref, instance, schema):
     return swagger_type_validator
 
 
+def type_validator(validator, types, instance, schema):
+    """Swagger 1.2 supports parameters of 'type': 'File'. Skip validation of
```
jnb: If you don't special-case the 'File' validation, and instead just use the general-case validation, does it fail?

pyramid_swagger/load_schema.py
```diff
@@ -124,6 +133,7 @@ def get_body_validator(models):
     {
         'paramType': ignore,
         'name': ignore,
+        'type': type_validator,
```
jnb: I can't see where this field is used?

pyramid_swagger/load_schema.py
analogue: Yes, ```test_200_with_file_upload(test_app)``` fails unless the special case is handled.

pyramid_swagger/load_schema.py
analogue: https://github.com/striglia/pyramid_swagger/blob/master/pyramid_swagger/load_schema.py#L146

jnb:
:+1:

analogue:
This is ready to merge. I updated the change log but left the version unchanged because I have another PR (separate bug) that needs to make it into the next release candidate.

PR #117, opened 2015-05-22
Fix request validation errors to surface as HTTP 4XX instead of HTTP5XX

analogue:
Before this change, a SwaggerMappingError originating from bravado-core during request validation would surface as HTTP 5XX errors instead of HTTP 4XX.  This PR updates the `validation_error` decorator to transform SwaggerMappingError (it already did this for jsonschema.exceptions.ValidationError).

The same will be done for responses during response validation (see `swaggerize_response`)


jnb:
:+1:

PR #118, opened 2015-05-26
Bump to 2.0.0-rc2

analogue:
This will fail in travis until https://github.com/Yelp/bravado-core/pull/21 is merged and bravado-core 1.0.0-rc1 is packaged and released.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,14 +36,10 @@ setup(
         'Programming Language :: Python :: 3.4',
     ],
     keywords='pyramid swagger validation',
-
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
-    package_data={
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -36,14 +36,10 @@
         'Programming Language :: Python :: 3.4',
     ],
     keywords='pyramid swagger validation',
-
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
-    package_data={
-        'pyramid_swagger': ['pyramid_swagger/swagger_spec_schemas/v1.2/*'],
-    },
```
analogue: FYI, package_data should have been removed when we switched over to using swagger_spec_validator.

striglia: ++ good catch

striglia:
Looks good once Bravado is released and the Travis run works

PR #119, opened 2015-05-28
Add pyramid_swagger.use_models config key

analogue:
This adds a new config key to pyramid_swagger and passes it on to bravado-core.  When ```use_models``` is set to ```False```, models are not generated from the incoming request. This retains compatibility with existing Swagger 1.2 views where models are represented as dicts.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -43,6 +43,11 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: True
         pyramid_swagger.enable_path_validation = True
 
+        # Use Python classes instead of dicts to represent models in incoming
+        # requests.
+        # Default: Whatever the default is in bravado_core, currently True
+        pyramid_swagger.use_models = True
+
         # Exclude certain endpoints from validation. Takes a list of regular
         # expressions.
         # Default: [r'^/static/?', r'^/api-docs/?, r'^/swagger.json']
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -45,8 +45,8 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
 
         # Use Python classes instead of dicts to represent models in incoming
         # requests.
-        # Default: Whatever the default is in bravado_core, currently True
-        pyramid_swagger.use_models = True
+        # Default: False
+        pyramid_swagger.use_models = False
 
         # Exclude certain endpoints from validation. Takes a list of regular
         # expressions.
--- a/docs/migrating_to_swagger_20.rst
+++ b/docs/migrating_to_swagger_20.rst
@@ -3,13 +3,12 @@ Migrating to Swagger 2.0
 
 So you're using pyramid_swagger with Swagger 1.2 and now it is time to upgrade to Swagger 2.0.
 
-Just set the version of Swagger to support via configuration. Also, set ``use_models`` to ``false`` so that your existing Swagger 1.2 views will continue to received models in incoming requests as dicts instead of Python classes.
+Just set the version of Swagger to support via configuration.
 
 .. code-block:: ini
 
         [app:main]
         pyramid_swagger.swagger_versions = ['2.0']
-        pyramid_swagger.use_models = false
 
 If you would like to continue servicing Swagger 1.2 clients, pyramid_swagger has you covered.
 
@@ -17,7 +16,6 @@ If you would like to continue servicing Swagger 1.2 clients, pyramid_swagger has
 
         [app:main]
         pyramid_swagger.swagger_versions = ['1.2', '2.0']
-        pyramid_swagger.use_models = false
 
 .. note::
 
@@ -31,7 +29,6 @@ If you're not using an ini file, configuration in Python also works.
     def main(global_config, **settings):
         # ...
         settings['pyramid_swagger.swagger_versions'] = ['2.0']
-        settings['pyramid_swagger.use_models'] = False
         # ...and so on with the other settings...
         config = Configurator(settings=settings)
         config.include('pyramid_swagger')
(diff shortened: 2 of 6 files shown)


tests/ingest_test.py
```diff
@@ -52,7 +53,6 @@ def test_get_swagger_schema_no_validation():
     get_swagger_schema(settings)
 
 
-@pytest.mark.xfail(reason='Remove 1.2 test')
```
analogue: Fixed this disabled test while I was in there

docs/migrating_to_swagger_20.rst
```diff
@@ -3,12 +3,13 @@ Migrating to Swagger 2.0
 
 So you're using pyramid_swagger with Swagger 1.2 and now it is time to upgrade to Swagger 2.0.
 
-Just set the version of Swagger to support via configuration like so:
+Just set the version of Swagger to support via configuration. Also, set ``use_models`` to ``false`` so that your existing Swagger 1.2 views will continue to received models in incoming requests as dicts instead of Python classes.
```
striglia: The purpose and effect of this option are unclear to me. Is "use_models" going to modify the data bound to my request? Should this read "existing Pyramid views" not "Swagger 1.2"...if not what is the Swagger view being referred to?

docs/configuration.rst
```diff
@@ -43,6 +43,11 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: True
         pyramid_swagger.enable_path_validation = True
 
+        # Use Python classes instead of dicts to represent models in incoming
```
striglia: Why do I (the app developer) care?

docs/configuration.rst
```diff
@@ -43,6 +43,11 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
...
+        # requests.
+        # Default: Whatever the default is in bravado_core, currently True
```
striglia: Can we explicitly default this inside pyramid_swagger code (e.g. if it's unset, set it false and pass that explicitly into bravado_core)? I wouldn't want this default to secretly change on users without an obvious code change here.

docs/migrating_to_swagger_20.rst
analogue: An example may better illustrate the behavior here.

With ```use_models=False```:

```python
@view_config(route_name='add.user)
def add_user(request):
    user = request.swagger_data['user']
    assert isinstance(user, dict)
    firstname = user['first_name']
    
    ...
```

With ```use_models=True```:
```python
@view_config(route_name='add.user)
def add_user(request):
    user = request.swagger_data['user']
    assert isinstance(user, bravado_core.models.User)
    firstname = user.first_name
    ...
```

No data bound to the request is changed. Only the representation of ```#/definitions``` in ```request.swagger_data``` is affected.

I'll make the suggested change to use "existing Pyramid views".

docs/configuration.rst
analogue: I hope the code sample in the previous comment makes this apparent.

docs/configuration.rst
analogue: Sure thing.

docs/migrating_to_swagger_20.rst
dnephin: Maybe add this example to the docs as well? I think there is a section for swagger_data already in quickstart

analogue: Done.

docs/configuration.rst
striglia: Yep I think it does :)

docs/quickstart.rst
```diff
@@ -79,6 +79,28 @@ Example:
 The raw values (not-casted to any type) are still available from their
 usual place on the request (`matchdict`, `GET`, `POST`, `json()`, etc)
 
+If you have ``pyramid_swagger.use_models`` enabled, you can interact with
+models defined in ``#/definitions`` as Python classes instead of dicts.
+
+.. code-block:: python
```
striglia: Can we include the snippet of the Swagger model here too? I think those three together will make it very clear what's going on.

docs/quickstart.rst
```diff
@@ -79,6 +79,28 @@ Example:
...
+
+    @view_config(route_name='add.user)
+    def add_user(request):
+        user = request.swagger_data['user']
```
striglia: Does it do casting too? Better to use a boolean type to make this crystal clear?

striglia:
Documentation improvements, but otherwise looks good!

docs/quickstart.rst
analogue: Done

docs/quickstart.rst
analogue: Good catch

docs/quickstart.rst
```diff
@@ -79,6 +79,48 @@ Example:
...
+            }
+          }
+        }
+      }
+      ...
+    }
+
+.. code-block:: python
+
+    @view_config(route_name='add.user)
```
prat0318: Nit: closing quote for add.user

docs/quickstart.rst
```diff
@@ -79,6 +79,48 @@ Example:
...
+        user = request.swagger_data['user']
+        assert isinstance(user, bravado_core.models.User)
+        first_name = user.first_name
+        ...
+
+Otherwise, models are represented as dicts.
+
+.. code-block:: python
+
+    @view_config(route_name='add.user)
```
prat0318: Same here.

docs/quickstart.rst
analogue: eagle eyes!

docs/quickstart.rst
analogue: done

striglia:
:+1: Let's do it. When we roll out the rc, we might check in to see if anyone is using the models flag yet and, if so, whether it still works the same with this new backend.

Issue #121, opened 2015-06-16
Error responses are incorrectly validated against the normal response type


chriskuehl:
Hello!

When response validation is enabled, error responses are validated against the normal response type, often producing errors.

For example, I have an endpoint defined like:
```json
{
    "swaggerVersion": "1.2",
    "apiVersion": "1.0",
    "basePath": "http://localhost",

    "apis": [
        {
            "path": "/return_an_int",
            "operations": [
                {
                    "method": "GET",
                    "nickname": "return_an_int",
                    "parameters": [],
                    "type": "integer",
                    "responseMessages": [
                        {
                            "code": 404,
                            "message": "Could not find an int"
                        }
                    ]
                }
            ]
        }
    ]
}
```

If my endpoint returns a 404 error, pyramid_swagger intercepts it and I end up with a 500 instead:

```html
HTTP/1.1 500 Internal Server Error
Content-Type: text/html; charset=UTF-8

<html>
 <head>
  <title>500 Internal Server Error</title>
 </head>
 <body>
  <h1>500 Internal Server Error</h1>
  The server has either erred or is incapable of performing the requested operation.<br/><br/>
u'Could not find an int.' is not of type 'integer'

Failed validating u'type' in schema:
    {u'type': 'integer'}

On instance:
    u'Could not find an int.'


 </body>
</html>
```

I traced this to [tween.py#L450](https://github.com/striglia/pyramid_swagger/blob/74479b4812c7052f534930edd7915704afcc0c2d/pyramid_swagger/tween.py#L450), where the response body is turned into either text or JSON, then passed to the validator. By the time the validator sees the content, it will have no idea what the status code is, so will be unable to properly validate the response.

An easy solution might be to not validate error responses at all, but I believe Swagger actually allows to specify types for error messages as well, so perhaps a correct fix would take those into account.

Many thanks! (Tested in v1.5.0-rc2.)

chriskuehl:
It seems that handling of errors is not super-well-defined in Swagger 1.2; see e.g. swagger-api/swagger-core#474, so trying to validate errors seems likely to be troublesome. Skipping validation for error responses, at least in Swagger 1.2, seems reasonable. Any thoughts?

striglia:
Hi Chris, thanks for the clear writeup :)

Agreed that response validation is pretty unclear in 1.2. As a general note, I've found that the 2.0 spec is wildly improved in most areas.

[...]

PR #122, opened 2015-06-16
Don't attempt to validate error responses

chriskuehl:
When response validation is enabled, error responses are validated against the normal response type, often producing errors. Instead, we should not validate error responses, as in Swagger 1.2, the types of error messages are not well defined.

I'm not able to create a PR against 1.5.0 because there is no branch on GitHub for it; instead, see the comparison against the 1.5.0-rc2 tag for a usefull diff: https://github.com/striglia/pyramid_swagger/compare/striglia:v1.5.0-rc2...chriskuehl:dont-validate-error-responses

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -352,6 +352,11 @@ def validate_response(response, validator):
         response.body in (None, b'', b'{}', b'null')
     ):
         return
+
+    # Don't attempt to validate non-success responses
+    if not 200 <= response.status_code <= 203:
+        return
+
     validator.validate(prepare_body(response))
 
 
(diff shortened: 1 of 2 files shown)


chriskuehl:
This fixes #121

striglia:
Awesome. I've just created a 1.2 branch for any ongoing maintenance releases we make like this. https://github.com/striglia/pyramid_swagger/tree/swagger_12

If you don't mind, let's rebase this PR on top of that branch and start from there. Once we commit this fix, are you ok working off the `swagger_12` branch, or will you need a proper release of the package?

chriskuehl:
thanks so much for the quick responses! opened #123 for this, unfortunately can't switch the branches involved in a PR without making a new one :(

PR #123, opened 2015-06-17
Don't attempt to validate error responses

### LINKED ISSUE #121: Error responses are incorrectly validated against the normal response type

chriskuehl:
When response validation is enabled, error responses are validated against the normal response type, often producing errors. Instead, we should not validate error responses, as in Swagger 1.2, the types of error messages are not well defined. This fixes #121.

This is a rebase of #122 onto the new swagger_12 branch. GitHub doesn't let you switch the branches involved in a pull request (boo) so here's a new PR.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+1.5.1 (2015-06-17)
+++++++++++++++++++++++
+
+* No longer attempts to validate error responses, which typically don't follow
+  the same format as successful responses. (Closes: #121)
+
 1.5.0 (2015-05-12)
 ++++++++++++++++++++++
 
--- a/pyramid_swagger/__about__.py
+++ b/pyramid_swagger/__about__.py
@@ -13,7 +13,7 @@ __title__ = "pyramid_swagger"
 __summary__ = "Swagger tools for use in pyramid webapps"
 __uri__ = "https://github.com/striglia/pyramid_swagger"
 
-__version__ = "1.5.0"
+__version__ = "1.5.1"
 
 __author__ = "Scott Triglia"
 __email__ = "scott.triglia@gmail.com"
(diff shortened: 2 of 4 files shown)


striglia:
Great :) From the sound of it, you'll want to make a new release available, so let's also have you bump setup.py's version to 1.5.1 and add an entry in the appropriate place to docs/changelog.rst. After that (and merging this PR) I'll tag v1.5.1 and upload it to pypi.

chriskuehl:
added a bump commit, thanks again!

PR #129, opened 2015-08-05
Add SwaggerFormatter integration.

### LINKED ISSUE #50: Support formats on types
e.g. date-time formatting. This is a semi-builtin feature of the jsonschema library, so it should be quite doable. See [this man entry](http://python-jsonschema.readthedocs.org/en/latest/validate/#validating-formats) for more information.

prat0318:
- Fixes #50 
- Have not bumped the version in this commit yet. Made an entry in CHANGELOG though.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,10 +1,10 @@
 Changelog
 =========
 
-2.X.X (2015-0X-XX)
+2.1.0 (2015-08-06)
 ++++++++++++++++++++++
 * Added ``user_formats`` configuration option to provide user-defined formats which can be used for validations
-  and conversions to wire-python-wire formats
+  and conversions to wire-python-wire formats. (See :ref:`user-format-label`)
 
 2.0.0 (2015-06-25)
 ++++++++++++++++++++++
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.X.X (2015-0X-XX)
+++++++++++++++++++++++
+* Added ``user_formats`` configuration option to provide user-defined formats which can be used for validations
+  and conversions to wire-python-wire formats
+
 2.0.0 (2015-06-25)
 ++++++++++++++++++++++
 * Added ``use_models`` configuration option for Swagger 2.0 backwards compatibility with existing pyramid views
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -130,6 +130,23 @@ After defining this format, it can be used in the Swagger Spec definition like s
             "format": "base64"
         }
 
+.. note::
+
+    The ``type`` need not be ``string`` always. The feature also works for other primitive
+    types like integer, boolean, etc. More details are in the Swagger Spec v2.0 `Data Types`_.
+
+    There are two types of validations which happen for user-defined formats.
+    The first one is the usual type checking which is similarly done for all the other values.
+    The second check is done by the ``validate`` function (provided by the user)
+    which is run on the serialised format. If the value doesn't conform to the format, the
+    ``validate`` function should raise :class:`bravado_core.exception.SwaggerValidationError`.
+
+    All the parameters to ``SwaggerFormat`` are mandatory. If you want any of the functions
+    to behave as a no-op, assign them a value ``lambda x: x``. On providing a user-format, the
+    default marshal/unmarshal behavior associated with that primitive type gets overridden by
+    the ``to_wire``/``to_python`` behavior registered with that user-format, respectively.
+
+.. _Data Types: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#user-content-data-types
 
 generate_resource_listing (Swagger 1.2 only)
 --------------------------------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -71,7 +71,8 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # below for more details
         pyramid_swagger.generate_resource_listing = false
 
-Note that, equivalently, you can add these during webapp configuration:
+
+Note that, equivalently, you can add these settings during webapp configuration:
 
 .. code-block:: python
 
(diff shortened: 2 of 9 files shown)


prat0318:
Checked manually running by it on `example_happyhour`

pyramid_swagger/tween.py
```diff
@@ -9,6 +9,7 @@
 
 import bravado_core
 from bravado_core.exception import SwaggerMappingError
+from bravado_core.formatter import SwaggerFormat  # noqa
```
prat0318: This is done so that the user can import SwaggerFormat from `pyramid_swagger` instead of importing `bravado-core` and importing from there.

docs/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.X.X (2015-0X-XX)
```
striglia: We can just make this 2.1.0.

Can you modify setup.py too?

docs/changelog.rst
```diff
@@ -1,6 +1,11 @@
...
+++++++++++++++++++++++
+* Added ``user_formats`` configuration option to provide user-defined formats which can be used for validations
+  and conversions to wire-python-wire formats
```
striglia: Link in the configuration explanation here

docs/configuration.rst
```diff
@@ -84,6 +85,49 @@ Note that, equivalently, you can add these during webapp configuration:
...
+After defining this format, it can be used in the Swagger Spec definition like so:
+
+.. code-block:: json
+
+        {
+            "name": "petId",
+            "in": "path",
+            "description": "ID of pet to return",
+            "required": true,
+            "type": "string",
```
striglia: Is "string" required as the type? Can other types work with custom formats?

docs/configuration.rst
```diff
@@ -84,6 +85,49 @@ Note that, equivalently, you can add these during webapp configuration:
...
+            settings['pyramid_swagger.user_formats'] = [user_format]
+
+
+And, ``user_format`` used above can be defined like this:
+
+.. code-block:: python
+
+        import base64
+        from pyramid_swagger.tween import SwaggerFormat
+        user_format = SwaggerFormat(format='base64',
```
striglia: Can we hyperlink in some documentation/source code for SwaggerFormat? Presumably we need to cross-link something from bravado core? If I'm a developer, I'm going to want to know more about the values these kwargs need to have and which are/aren't optional.

tests/tween_test.py
```diff
@@ -234,6 +236,14 @@ def test_get_swagger_versions_unsupported():
     assert 'Swagger version 10.0 is not supported' in str(excinfo.value)
 
 
+def test_user_defined_formats_get_registered():
```
striglia: Can we add at least one acceptance test? This is pretty custom functionality, so it'd be good to stay confident it's actually working end to end. Maybe a positive and negative test?

docs/changelog.rst
prat0318: ack.

docs/configuration.rst
prat0318: Yup, any primitive type should work.

docs/changelog.rst
prat0318: ack.

docs/configuration.rst
prat0318: The only link i can think of is a bravado-core github link for that [snippet](https://github.com/Yelp/bravado-core/blob/master/bravado_core/formatter.py#L72). User can anytime fetch the docstring of SwaggerFormat though.

tests/tween_test.py
prat0318: Yeah ok, will see how easy it is to add one currently.

docs/configuration.rst
dnephin: +1 

We should fix these docs http://bravado-core.readthedocs.org/en/latest/bravado.html so it actually contains all the classes/functions, then you can link to it using ":class:`bravado_core.SwaggerFormat`" (or whatever the import path is)

docs/configuration.rst
striglia: Oh interesting. So I should think of user_formats as *additional* verification on top of the normal primitive verification? Clarifying the interplay between any inherent validation and the user_format validation in the documentation section would be good.

This feature is going to be pretty powerful, so we should make sure all that power is clearly explained here :)

prat0318: Yeah, it does [both](https://github.com/Yelp/bravado-core/blob/master/bravado_core/validate.py#L52,L53) of the validations one after the other.

tests/tween_test.py
prat0318: added happy and failure cases.

docs/configuration.rst
```diff
@@ -84,6 +85,52 @@ Note that, equivalently, you can add these during webapp configuration:
...
+
+``user_format`` used above is an instance of
+:class:`bravado_core.formatter.SwaggerFormat` (`ref <http://bravado-core.readthedocs.org/en/latest/bravado_core.html#bravado_core.formatter.SwaggerFormat>`_) and can be defined like this:
+
+.. code-block:: python
+
+        import base64
+        from pyramid_swagger.tween import SwaggerFormat
+        user_format = SwaggerFormat(format='base64',
+                                    to_wire=base64.b64encode,
```
striglia: How should pyramid_swagger users expect the to_python and to_wire arguments to be used in practice? Are they required or is the fallback behavior defined? How do they interact with the default SerDe behavior of the primitive type we're describing?

docs/configuration.rst
striglia: Cool! I'd call this out explicitly in this documentation.

tests/acceptance/format_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+@pytest.fixture
+def test_app_with_base64(settings, user_format):
+    """Fixture for setting up a Swagger 2.0 version of the test test_app."""
+    settings['pyramid_swagger.swagger_versions'] = ['2.0']
+    settings['pyramid_swagger.user_formats'] = [user_format]
+    return TestApp(main({}, **settings))
+
+
+def test_user_format_happy_case(test_app_with_base64):
```
striglia: Nice tests :)

docs/configuration.rst
prat0318: ack. will add that.

docs/configuration.rst
prat0318: Added in the new commit.

docs/configuration.rst
```diff
@@ -84,6 +85,69 @@ Note that, equivalently, you can add these during webapp configuration:
...
+        }
+
+.. note::
+
+    The ``type`` need not be ``string`` always. The feature also works for other primitive
+    types like integer, boolean, etc. More details are in the Swagger Spec v2.0 `Data Types`_.
+
+    There are two types of validations which happen for user-defined formats.
+    The first one is the usual type checking which is similarly done for all the other values.
+    The second check is done by the ``validate`` function (provided by the user)
```
striglia: "``validate`` function (from the user_format you configured for this type)? Took me a minute to figure out where the ``validate`` function was being provided.

docs/configuration.rst
```diff
@@ -84,6 +85,69 @@ Note that, equivalently, you can add these during webapp configuration:
...
+    which is run on the serialised format. If the value doesn't conform to the format, the
+    ``validate`` function should raise :class:`bravado_core.exception.SwaggerValidationError`.
```
striglia: should or must?

striglia:
Just a couple documentation comments. I think this one is otherwise ready to go :)

docs/configuration.rst
prat0318: `should`. If any other error is raised, `bravado-core` wraps it up to this error.

docs/configuration.rst
prat0318: ack.

docs/configuration.rst
striglia: Cool! Then it sounds like it must raise an error, and that error should be bravado_core's SwaggerValidationError.

prat0318: ack. will modify the wordings.

striglia:
All looks good to me. Anything left to add before we merge?

We can also fold in a bugfix or two before we tag 2.1.0.

prat0318:
Nothing else i can think of. We would need to add #131 soon, i am open to add that enhancement as part of 2.2.0

striglia:
Let's just wait on 2.1.0 until we include it. No rush.

PR #132, opened 2015-08-11
Send absolute swagger spec url to bravado-core

### LINKED ISSUE #131: Send origin_uri across to bravado_core
Send absolute address of `swagger.json` to bravado_core so that cross-ref `$ref` can be dereferenced even if they have a relative path inside, like: `$ref: "parameters/foo.json#/bar"`

This would need the schema_url [here](https://github.com/striglia/pyramid_swagger/blob/master/pyramid_swagger/ingest.py#L175)

prat0318:
- Send absolute url to `Spec.from_dict` so that 'relative' `$ref`  get resolved relative to the absolute url of `swagger.json` spec.
- Fixes #131

### DIFF
--- a/pyramid_swagger/ingest.py
+++ b/pyramid_swagger/ingest.py
@@ -6,6 +6,7 @@ import os.path
 
 import simplejson
 from bravado_core.spec import Spec
+from six.moves.urllib import parse as urlparse
 
 from .api import build_swagger_12_endpoints
 from .load_schema import load_schema
@@ -167,12 +168,16 @@ def get_swagger_spec(settings):
     :rtype: :class:`bravado_core.spec.Spec`
     """
     schema_dir = settings.get('pyramid_swagger.schema_directory', 'api_docs/')
-    with open(os.path.join(schema_dir, 'swagger.json'), 'r') as f:
+    schema_path = os.path.join(schema_dir, 'swagger.json')
+    schema_url = urlparse.urljoin('file:', os.path.abspath(schema_path))
+
+    with open(schema_path, 'r') as f:
         spec_dict = simplejson.loads(f.read())
 
     return Spec.from_dict(
         spec_dict,
-        config=create_bravado_core_config(settings))
+        config=create_bravado_core_config(settings),
+        origin_url=schema_url)
 
 
 def create_bravado_core_config(settings):
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,10 +1,11 @@
 Changelog
 =========
 
-2.1.0 (2015-08-06)
+2.1.0 (2015-08-14)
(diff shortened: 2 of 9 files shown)


pyramid_swagger/ingest.py
```diff
@@ -167,12 +168,16 @@ def get_swagger_spec(settings):
     :rtype: :class:`bravado_core.spec.Spec`
     """
     schema_dir = settings.get('pyramid_swagger.schema_directory', 'api_docs/')
-    with open(os.path.join(schema_dir, 'swagger.json'), 'r') as f:
+    schema_path = os.path.join(schema_dir, 'swagger.json')
```
striglia: Do we have a ticket for YAML support? Would be nice to throw a comment here linking to a github issue

pyramid_swagger/ingest.py
```diff
@@ -167,12 +168,16 @@ def get_swagger_spec(settings):
...
+    schema_url = urlparse.urljoin('file:', os.path.abspath(schema_path))
+
+    with open(schema_path, 'r') as f:
         spec_dict = simplejson.loads(f.read())
 
     return Spec.from_dict(
         spec_dict,
-        config=create_bravado_core_config(settings))
+        config=create_bravado_core_config(settings),
+        origin_url=schema_url)
```
striglia: Any detectable behavior we expect from our side? Should we add an internal reference to our test schemas to exercise this?

pyramid_swagger/ingest.py
prat0318: The relevant ticket is Yelp/bravado-core#42. Do you want the comment in the code as well, that should not be necessary?

pyramid_swagger/ingest.py
prat0318: We can add a cross-ref example as an acceptance test. This would need bumping up the bravado-core version with this patch. I will try to add that test.

striglia:
We'll wait on this before releasing 2.1.0

pyramid_swagger/ingest.py
prat0318: Added the tests.

pyramid_swagger/ingest.py
striglia: Nah, this is probably good for now.

pyramid_swagger/ingest.py
striglia: Nice! Those are definitely some complicated relative refs :)

striglia:
Looks good. And the new tests failed before your code change?

Can we add the relevant changelog entry?

prat0318:
Yes, they will give a `JsonRef` error on older bravado-core version OR `foo.json : Not found` without this patch.

PR #136, opened 2015-08-19
Add include_2dot0_routes config option.

### LINKED ISSUE #135: Add a new config to include some specific 2.0 endpoints
Currently, incremental migration of endpoints from v1.2 to v2.0 is not possible. If `swagger_versions:['1.2','2.0']` is present in the config and `swagger.json` is present (which is very small right now and contains only a single migrated endpoint), then `pyramid_swagger` will look for all other endpoints in the same `swagger.json` and hence fail.

The proposed solution is to add a new config :  
```
pyramid_swagger.include_2dot0_routes = ['endpoint_route1']
```
And IF `2.0` is in `swagger_versions` AND `include_2dot0_routes` is non-empty AND current route is in `include_2dot0_routes`, only then go for `2.0` swagger_handler otherwise use the `1.2` swagger handler.

prat0318:
Fixes #135


pyramid_swagger/tween.py
```diff
@@ -93,6 +99,28 @@ def _get_validation_context(registry):
...
+    Exception is made when `settings.include_2dot0_routes` are non-empty and
+    ['1.2', '2.0'] both are present in available swagger versions. In this
+    special scenario, '2.0' spec is chosen only for requests which are listed
+    in the `include_2dot0_routes`. This helps in incremental migration of
+    routes from v1.2 to v2.0 by making moving to v2.0 opt-in.
+
+    :rtype: :class:`SwaggerHandler`
+    """
+    enabled_swagger_versions = get_swagger_versions(registry.settings)
+    if (SWAGGER_20 in enabled_swagger_versions
```
prat0318: these conditions may need some explanations. This special condition is checked if both v1.2 and v2.0 versions are enabled. If no `include_2dot0_routes` is passed, then this special case should not be checked (this will be the usual flow). If `include..routes` is non-empty then check if given route is not present in the included routes, if they are not, use the v1.2 flow for this request.

pyramid_swagger/tween.py
```diff
@@ -273,6 +301,8 @@ def load_settings(registry):
     return Settings(
         schema=schema,
         swagger_handler=build_swagger_handler(registry.settings, schema),
+        swagger12_handler=build_swagger12_handler_wrapper(
```
prat0318: This caching is done so that for every request no new swagger12_handler is created. This will be `None` if v1.2 is not in the enabled swagger versions. This will be duplicated in case only enabled swagger version was v1.2 but i think it is OK as it simplifies the code.

pyramid_swagger/tween.py
```diff
@@ -302,24 +334,42 @@ def build_swagger_handler(settings, schema):
...
+
+def build_swagger12_handler(schema):
+    return SwaggerHandler(
+        op_for_request=schema.validators_for_request,
+        handle_request=handle_request,
+        handle_response=validate_response,
+    )
+
+
+def build_swagger12_handler_wrapper(schema):
```
prat0318: This wrapper is needed to check that schema is non-null. It could be `None` if swagger_versions only had `2.0` and `build_swagger12_handler_wrapper` is called. It should simply return `None` in this case.

pyramid_swagger/tween.py
```diff
@@ -93,6 +99,28 @@ def _get_validation_context(registry):
...
+    """
+    enabled_swagger_versions = get_swagger_versions(registry.settings)
+    if (SWAGGER_20 in enabled_swagger_versions
+        and SWAGGER_12 in enabled_swagger_versions
+        and settings.include_2dot0_routes
+        and route_info.get('route')
+            and route_info['route'].name not in settings.include_2dot0_routes):
+        return settings.swagger12_handler
+
+    return settings.swagger_handler
```
striglia: This strikes me as really confusing. It's not obvious that settings.swagger_handler might actually be equivalent to settings.swagger12_handler.

Can we just store them clearly as swagger12_handler and swagger20_handler, then use all the configuration together for validation in this method? Splitting the logic for choosing what `settings.swagger_handler` is between this method and `build_swagger_handler` is pretty surprising and error prone.

pyramid_swagger/tween.py
```diff
@@ -42,18 +42,21 @@ class Settings(namedtuple(
     [
```
striglia: Going to definitely need some updated docs+changelog on this one. Can wait until we've finalized the implementation.

pyramid_swagger/tween.py
prat0318: yeah, i thought about it. Current approach just saves those two `if` checks from happening with every request. I am open to move those checks inside this method.

striglia: I'd vote for code simplicity over reducing the number of if statements per call. I bet our jsonschema serde cost swamps it :)

prat0318: Added [new commit](https://github.com/prat0318/pyramid_swagger/commit/96bac57ccfb551a0748ba1df2f68de3f467e2c9a) to separate out swagger handlers and define them explicitly.

pyramid_swagger/tween.py
```diff
@@ -41,19 +41,22 @@ class Settings(namedtuple(
...
         'schema',
-        'swagger_handler',
+        'swagger12_handler',
+        'swagger20_handler',
         'validate_request',
         'validate_response',
         'validate_path',
         'exclude_paths',
         'exclude_routes',
+        'include_2dot0_routes',
```
striglia: "prefer"? And we seem to have a pattern of abbreviating 2.0 as "20" elsewhere, so I'd be consistent here

pyramid_swagger/tween.py
```diff
@@ -93,6 +99,34 @@ def _get_validation_context(registry):
         return noop_context
 
 
+def get_swagger_handler(settings, route_info, registry):
+    """Contains callables that isolate implementation differences in the tween
+    to handle both Swagger 1.2 and Swagger 2.0.
+
+    Exception is made when `settings.include_2dot0_routes` are non-empty and
```
striglia: I'd probably move this down to a block comment on the if statement.

prat0318: I think it is helpful to keep it in the method docstring. (comments inside the method wont reflect)

pyramid_swagger/tween.py
prat0318: ok, sg2m

pyramid_swagger/tween.py
striglia: Works for me

prat0318:
Added unit tests for all cases and a docs section.

docs/migrating_to_swagger_20.rst
```diff
@@ -40,3 +40,23 @@ Next, create a Swagger 2.0 version of your swagger schema. There are some great
...
+.. code-block:: ini
+
+        [app:main]
+        pyramid_swagger.prefer_20_routes = ['route_foo']
+
+.. note::
+
+    The above config is read only when both `['1.2', '2.0']` are present in `swagger_versions` config. If that
+    is the case and the request's route is present in `prefer_20_routes`, ONLY then the request is served through
+    swagger 2.0 otherwise through 1.2. The only exception is when the config is not defined at all, in that case
```
striglia: "the config is not defined at all"? Meaning this particular option is not set? This isn't the only exception right...you could also not specity 1.2 and 2.0?

tests/acceptance/include_config_test.py
```diff
@@ -0,0 +1,49 @@
...
+    return {
+        'pyramid_swagger.schema_directory': dir_path,
+        'pyramid_swagger.enable_request_validation': True,
+        'pyramid_swagger.enable_swagger_spec_validation': True,
+        'pyramid_swagger.swagger_versions': ['1.2', '2.0'],
+    }
+
+
+@pytest.fixture
+def test_app_with_no_include_conf(settings):
```
striglia: "include conf" or "include_config" is the old name. Think this module needs an update

tests/acceptance/include_config_test.py
```diff
@@ -0,0 +1,49 @@
...
+    """Fixture for setting up a Swagger 2.0 version of the test test_app."""
+    return TestApp(main({}, **settings))
+
+
+@pytest.fixture
+def test_app_with_include_conf(settings):
+    """Fixture for setting up a Swagger 2.0 version of the test test_app."""
```
striglia: These docstrings are a little stale/wrong/general

tests/sample_schemas/include_config_app/api_docs.json
```diff
@@ -0,0 +1,9 @@
+{
```
striglia: Same comment on all these uses of "include_config" in the filepaths

docs/migrating_to_swagger_20.rst
prat0318: will add that bit.

prat0318:
Renamed modules.

striglia:
Awesome. I think we're set to merge and tag.

Issue #120, opened 2015-06-07
Pyramid view configuration could not register the view route


yapengwu:
When I try the pets swagger specification with simple pyramid application. I could not get view configuration working. I keep get the following error:
**pyramid.exceptions.ConfigurationExecutionError: <class 'pyramid.exceptions.ConfigurationError'>: No route named api.pets.get found for view registration**

Here are the pyramid configuration, view configuration and swagger example I used.
__init__.py:
```
from pyramid.config import Configurator


def main(global_config, **settings):
    """ This function returns a Pyramid WSGI application.
    """
    settings['pyramid_swagger.swagger_versions'] = ['2.0']
    config = Configurator(settings=settings)
    config.include('pyramid_chameleon')
    config.include('pyramid_swagger')
    config.add_static_view('static', 'static', cache_max_age=3600)
    config.add_route('home', '/')
    config.scan()
    return config.make_wsgi_app()
```

views.py:

```
from pyramid.view import view_config


@view_config(route_name='home', renderer='templates/mytemplate.pt')
def my_view(request):
    return {'project': 'mypramid'}

@view_config(route_name='api.pets.get')
def get_pets(request):
    # Returns thing_id as an int (assuming the swagger type is integer)
    pets_id = request.swagger_data['id']
    return {'project': 'mypramid'}
```

api_docs/swagger.json:
```
{
    "swagger": "2.0",
    "info": {
        "version": "1.0.0",
        "title": "Swagger Petstore (Simple)",
        "description": "A sample API that uses a petstore as an example to demonstrate features in the swagger-2.0 specification",
        "termsOfService": "http://helloreverb.com/terms/",
        "contact": {
            "name": "Swagger API team",
            "email": "foo@example.com",
            "url": "http://swagger.io"
        },
        "license": {
            "name": "MIT",
            "url": "http://opensource.org/licenses/MIT"
        }
    },
    "host": "localhost:6543",
    "basePath": "/api",
    "schemes": [
        "http"
    ],
    "consumes": [
        "application/json"
    ],
    "produces": [
        "application/json"
    ],
    "paths": {
        "/pets": {
            "get": {
                "description": "Returns all pets from the system that the user has access to",
                "operationId": "findPets",
                "produces": [
                    "application/json",
                    "application/xml",
                    "text/xml",
                    "text/html"
                ],
                "parameters": [
                    {
                        "name": "tags",
                        "in": "query",
                        "description": "tags to filter by",
                        "required": false,
[...]

PR #141, opened 2015-09-24
Quickstart documentation update to eliminate confusion about routes.

brianthelion:
Routes aren't currently auto-populated. This is not obvious from the Quickstart documentation, leading to at least one reported issue (#120) and some mis-spent hours. This commit adds some clarifying language to docs/quickstart.rst.


striglia:
Thanks for the patch!

Issue #139, opened 2015-08-28
Support Python 3.x again


striglia:
We seem to have lost support for this during the 2.0.0 refactor that included bravado-core as a dependency.

See https://github.com/striglia/pyramid_swagger/blob/master/tox.ini#L2 for where we disabled testing.

We got a bug report that running 2to3 on the repo complained about this line:
```
File "/Users/.../.virtualenvs/.../lib/python3.4/site-packages/pyramid_swagger/tween.py", line 418
raise exc_class(str(exc)), None, sys.exc_info()[2]
```

Short term, we need to remove py3x trove classifiers from setup.py.
To complete this ticket, we need to:

* fix any existing non-3-compatible issues in the repo
* re-enable py3x testing in tox.ini
* put py3x trove classifiers back into setup.py

msander:
+1 for this

striglia:
Resolved in #149

PR #149, opened 2015-12-04
Py3X support

msander:


msander:
docs are failing because https://helloreverb.com/developers/swagger is dead

striglia:
Thanks for the PR!

You can definitely just replace the reverb link with http://swagger.io/getting-started/. I'm a little surprised using bravado-core isn't giving us any py3 issues....but I guess the proof is in the CI run.

Resolves #139

msander:
Bravado-core seems to support python 3 since version 1.1.0 (2015-06-25).
Nevertheless I haven't used it yet. I'm building my first project with swagger on python 3 and pyramid.
Any specific issues you have in mind which need to be looked at carefully?

striglia:
Nah, I am probably mis-remembering how this came up in the first place. I say we let the tests be the judge and we can always put out a bugfix release if that doesn't do it.

PR #150, opened 2016-02-04
This adds back support for pyramid 1.6

sontek:


sontek:
Pyramid 1.6 changed the default status code for unknown errors to `520`.  So I just updated the tests to allow that. The failing pep8 and docs wasn't caused by this PR, master fails as well.

djeebus:
Thanks @sontek! I was hoping this would show up soon =)

djeebus:

sontek:
@striglia I updated this with a cleaner fix.  The pep8 and docs are failing because of existing problems in master but it looks like @djeebus has sent PRs to fix all of them.

PR #152, opened 2016-02-04
Fix pep8 issues

djeebus:
I'm taking a shot at this to clean up the tests.


djeebus:

PR #153, opened 2016-02-04
Fix issue with parsing json

djeebus:
I removed the "..." that was causing the docs to fail to be built. The rest was just IntelliJ stripping the whitespace from the end of the lines.


djeebus:

PR #154, opened 2016-02-04
All fixes wrapped up

djeebus:
This wraps up #150, #152, and #153. This should ensure that the travis build passes.

### DIFF
--- a/pyramid_swagger/exceptions.py
+++ b/pyramid_swagger/exceptions.py
@@ -2,10 +2,10 @@
 import sys
 
 import six
-from pyramid.httpexceptions import HTTPClientError, HTTPInternalServerError
+from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
 
 
-class RequestValidationError(HTTPClientError):
+class RequestValidationError(HTTPBadRequest):
     pass
 
 
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -117,11 +117,14 @@ def get_swagger_objects(settings, route_info, registry):
     schema12 = registry.settings['pyramid_swagger.schema12']
     schema20 = registry.settings['pyramid_swagger.schema20']
 
-    if (SWAGGER_20 in enabled_swagger_versions
-        and SWAGGER_12 in enabled_swagger_versions
-        and settings.prefer_20_routes
-        and route_info.get('route')
-            and route_info['route'].name not in settings.prefer_20_routes):
+    fallback_to_swagger12_route = (
+        SWAGGER_20 in enabled_swagger_versions and
+        SWAGGER_12 in enabled_swagger_versions and
+        settings.prefer_20_routes and
+        route_info.get('route') and
+        route_info['route'].name not in settings.prefer_20_routes
+    )
+    if fallback_to_swagger12_route:
         return settings.swagger12_handler, schema12
 
     if SWAGGER_20 in enabled_swagger_versions:
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -552,7 +552,10 @@ def get_op_for_request(request, route_info, spec):
     # pyramid.urldispath.Route
     route = route_info['route']
     if hasattr(route, 'path'):
-        op = spec.get_op_for_request(request.method, route.path)
+        route_path = route.path
+        if route_path[0] != '/':
+            route_path = '/' + route_path
+        op = spec.get_op_for_request(request.method, route_path)
         if op is not None:
             return op
     raise PathNotMatchedError(
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 # TODO: re-enable py33,py34 when bravado-core is py3x compatible - https://github.com/Yelp/bravado-core/issues/10
-envlist = py26,py27,py33,py34,pep8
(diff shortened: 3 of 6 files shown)


striglia:
Thanks for both the fixes and pulling them all together. I'm aiming to block out some time this weekend to poke around the individual CRs.

I admit I'm a little confused about the build comments, latest build I see on master shows green: https://travis-ci.org/striglia/pyramid_swagger?branch=master. Maybe you're just referring to adding back in the pyramid 1.6 support?

djeebus:
@striglia thanks for building `pyramid_swagger` in the first place!

The travis build report doesn't say which version of pyramid it's installing, and I imagine that the virtualenv was created a long time ago, which means it's using an old version of pyramid. The bug caused by pyramid 1.6 is also caused by pyramid>1.5.4. If you add the `-r` flag to tox, I believe the bug will be recreated on `master`.

djeebus:
@striglia: @sontek cleaned up his fix for pyramid 1.5.4+ support, looks _much_ cleaner!

striglia:
Ah good eye. I had assumed Travis was blowing away environments between runs, but your explanation matches the symptoms here :) If @sontek doesn't mind, I'll probably just collect any comments in this PR for ease of reading.

sontek:
Sounds good to me!

pyramid_swagger/tween.py
```diff
@@ -117,11 +117,14 @@ def get_swagger_objects(settings, route_info, registry):
...
-        and settings.prefer_20_routes
-        and route_info.get('route')
-            and route_info['route'].name not in settings.prefer_20_routes):
+    fallback_to_swagger12_route = (
+        SWAGGER_20 in enabled_swagger_versions and
+        SWAGGER_12 in enabled_swagger_versions and
+        settings.prefer_20_routes and
+        route_info.get('route') and
+        route_info['route'].name not in settings.prefer_20_routes
+    )
```
sontek: This could look like this:

```python
if all([
    SWAGGER_20 in enabled_swagger_versions,
    SWAGGER_12 in enabled_swagger_versions,
    settings.prefer_20_routes,
    route_info.get('route'),
    route_info['route'].name not in settings.prefer_20_routes
]):
    # do stuff
```

pyramid_swagger/exceptions.py
```diff
@@ -2,10 +2,10 @@
 import sys
 
 import six
-from pyramid.httpexceptions import HTTPClientError, HTTPInternalServerError
+from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
 
 
-class RequestValidationError(HTTPClientError):
+class RequestValidationError(HTTPBadRequest):
```
striglia: I'm not sure I understand the difference between inheriting from HTTPBadRequest vs HTTPClientError. http://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html makes it look like maybe the former isn't implying an explicit return code? Mind helping me out here?

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 # TODO: re-enable py33,py34 when bravado-core is py3x compatible - https://github.com/Yelp/bravado-core/issues/10
-envlist = py26,py27,py33,py34,pep8
+envlist = py26,py27,py33,py34,py35,{py27,py35}-pyramid15,pep8
```
striglia: Wow very cool. So if I'm reading right, these are the factors referred to in http://testrun.org/tox/latest/config.html#factors-and-factor-conditional-settings and basically say "make a special pair of environments with py27 and py35 where we install an old version of pyramid"?

And I guess we rely on tox installing these deps before it installs the package itself to use this version over the setup.py one?

pyramid_swagger/exceptions.py
sontek: Yeah, `HTTPClientError` doesn't have a status code so it'll default to `520`.  The behavior you were relying on prior to pyramid 1.6 still exists but inheriting from `HTTPBadRequest`.  You could also just set the status codes manually if you prefer that.

tox.ini
sontek: Correct!  Since `deps` runs before the `installcommand` it'll already have pyramid installed when the tests run and wont pull it in:

```
$ .tox/py35-pyramid15/bin/pip freeze|grep pyramid
pyramid==1.5.4
pyramid-swagger==2.2.2


$ .tox/py35/bin/pip freeze|grep pyramid
pyramid==1.6.1
pyramid-swagger==2.2.2
```

pyramid_swagger/exceptions.py
striglia: Great. I prefer to just use the right exception class, although this is technically an interface change I'm really really hoping any users who care are directly catching RequestValidationError, not the inherited error class. 

Feel like something where a bugfix bump to the version with a clear CHANGELOG is sufficient warning?

sontek: Yeah, I think that should be sufficient.  If people were catching `HTTPClientError` they would have basically had a catch all exception handler because its the base class for pyramid http client exception.  But even with that, `HTTPBadRequest` subclasses it so their code should still work and shouldn't be considered an interface change

striglia:
merged, pushed, tagged as v2.2.3, and published to https://pypi.python.org/pypi/pyramid_swagger/. Thanks for the help and patience :smile:  If you have any larger issues or feedback, feel free to open up further issues on the project!

Issue #134, opened 2015-08-14
Support YAML specs for 2.0


striglia:
Pretty simple for pyramid_swagger. Need to get bravado-core to support it ([issue link](https://github.com/Yelp/bravado-core/issues/42)) and then update a small amount of code in pyramid_swagger to accept both json and yaml.

djeebus:
As a heads up, I added yaml support (in Yelp/bravado-core#73) to bravado-core, and I'm working on adding yaml support to pyramid-swagger. Progress is being made! =)

djeebus:
The branch is here if you're anxious to see what it looks like: https://github.com/djeebus/pyramid_swagger/tree/add-yaml-support

I'm waiting on bravado-core to release a new version that we can depend on in pyramid-swagger. *taps foot impatiently* =)

analogue:
@djeebus bravado-core 4.1.0 is up on pypi now

djeebus:
@analogue fantastic! =) Thanks, I'll get the yaml PR here soon.

striglia:
Resolved by #156

PR #156, opened 2016-03-02
Add yaml support

djeebus:
This was a bit more straightforward than I expected. After the changes in bravado-core, all I needed to do was reuse the bravado-core url handlers to open the file.

I also added a new configuration directive that allows you to define the file name (allowing you to use `swagger.yaml`). 

Two future ideas:
1.Ccombine `pyramid_swagger.schema_directory` and `pyramid_swagger.schema_file` to a single directive. The two fields seem redundant.
2. There are some issues with support for the `/api-docs/swagger.json` endpoint. It works on its own, but the yaml files that it links to are, well, yaml files, which prevents any swagger UIs from consuming them. I was considering finding a way to recombine the external yaml files into a single json file before returning it, but wanted to get people's feedback on that idea first.

Enjoy!

Edit: Solves #134

### DIFF
--- a/pyramid_swagger/__init__.py
+++ b/pyramid_swagger/__init__.py
@@ -4,7 +4,10 @@ Import this module to add the validation tween to your pyramid app.
 """
 import pyramid
 
-from .api import register_api_doc_endpoints, build_swagger_20_swagger_dot_json
+from .api import (
+    register_api_doc_endpoints,
+    build_swagger_20_swagger_schema_views,
+)
 from .ingest import get_swagger_schema
 from .ingest import get_swagger_spec
 from .tween import get_swagger_versions, SWAGGER_12, SWAGGER_20
@@ -16,6 +19,11 @@ def includeme(config):
     """
     settings = config.registry.settings
     swagger_versions = get_swagger_versions(settings)
+
+    # for rendering /swagger.yaml
+    config.add_renderer(
+        'yaml', 'pyramid_swagger.api.YamlRendererFactory',
+    )
 
     # Add the SwaggerSchema to settings to make it available to the validation
     # tween and `register_api_doc_endpoints`
@@ -45,5 +53,5 @@ def includeme(config):
         if SWAGGER_20 in swagger_versions:
             register_api_doc_endpoints(
                 config,
-                [build_swagger_20_swagger_dot_json(config)],
+                build_swagger_20_swagger_schema_views(config),
                 base_path='')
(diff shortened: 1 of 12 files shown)


striglia:
Sweet sweet yaml support :) Thanks for putting this together!

1) Seems like a solid plan for 3.0. The directory option is mostly a holdover from Swagger1.2 which likely should be entirely removed for 3.0 anyway.
2) Agreed that having multiple on-disk representations mean we need a consistent story on what precisely the HTTP API serves. Having discussed this before for cross-language services, my gut is that you should serve a complete, valid 2.0 spec over the endpoint. That basically means pyramid_swagger needs to compile your on-disk format (anything translatable to json) into the full spec and serve that up.

That make sense? Sound consistent?

djeebus:
Yup, makes sense to me. Do you want me to take care of compiling the yaml spec to a full json spec in this PR, or in a separate one?

striglia:
Hrm. It occurs to me that the OpenAPI spec actually allows YAML now as the direct representation:

```
The files describing the RESTful API in accordance with the Swagger specification are represented as JSON objects and conform to the JSON standards. YAML, being a superset of JSON, can be used as well to represent a Swagger specification file.
```

Now I guess this is more confusing than interesting, since they should be two ways to represent the same data structure once we've json.loaded or yaml.loaded them into Python objects.

If you don't mind, let's just do it all in here. I assume the current sticking point is the
```
"$ref": "defs.yaml#/definitions/standard_response"
```
line in your sample spec? Can we test the compiling of the single file easily...perhaps hitting the /api-docs view and confirming it contains data pulled in from a $ref to a yaml file?

djeebus:
Yup, I'll give that a shot. It'd be pretty easy to expose both endpoints (`/api-docs/swagger.yaml` and `/api-docs/swagger.json`) and compile down to the appropriate format.

djeebus:
@striglia I believe we're looking good now. Let me know if there's anything else I can do!

striglia:
@djeebus looks like you have some brief conflicts with master (probably in setup.py). Mind pulling and resolving?

djeebus:
@striglia done!

striglia:
Thanks as always :) I'm going to leave this unreleased for now and hopefully get some time to test it out in the wild. If you have any ability to do so as well, that'd be appreciated. Let me know if getting the new version tagged and released on PyPI becomes a priority for you and I'm happy to pull it together.

cc @sjaensch as I think he's working on this from the Bravado (client) side currently

lukas-b:
This is sweet, thanks @djeebus ! :)

sjaensch:
For bravado this was merged in Yelp/bravado#198, the issue being Yelp/bravado#168.

striglia:
This is out in 2.3.0-rc1 now (here in repo and in pypi). If all looks happy (and after at least the outstanding PR gets in) it'll probably be time for 2.3.0 proper.

PR #157, opened 2016-03-03
Only run tests in the `test/` folder.

djeebus:
This enables people to create their virtualenv in the same folder that 
the project is in, and still enable them to run `tox`. Without it, pytest
attempts to process all the files in your virtualenv.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ deps =
     pyramid15: pyramid<=1.5.4
 
 commands =
-    coverage run --source=pyramid_swagger/ --omit=pyramid_swagger/__about__.py -m pytest --capture=no --strict {posargs}
+    coverage run --source=pyramid_swagger/ --omit=pyramid_swagger/__about__.py -m pytest --capture=no --strict tests/ {posargs}
     coverage report -m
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -11,7 +11,7 @@ deps =
     pyramid15: pyramid<=1.5.4
 
 commands =
-    coverage run --source=pyramid_swagger/ --omit=pyramid_swagger/__about__.py -m pytest --capture=no --strict {posargs}
+    coverage run --source=pyramid_swagger/ --omit=pyramid_swagger/__about__.py -m pytest --capture=no --strict tests/ {posargs}
```
striglia: Pytest is always vigilant. Possibly a little too vigilant for my tastes :)

Instead of always including tests like this, lets make it the default value of posargs using http://tox.readthedocs.org/en/1.0/example/general.html#interactively-passing-positional-arguments.

That way you can always invoke like `tox -- tests/foobar_test.py` to avoid running the whole suite.

striglia:
:ship:

djeebus:
Thanks!

PR #159, opened 2016-03-17
Response/Request property access improvements

bwarfield-amplify:
This attempts to resolve an issue I encountered using a swagger spec that included header information in the response:

    "headers": {
        "Last-Modified": {
            "description": "The last date at which the results of your query would have changed",
            "type": "string",
            "format": "date-time"
        }
    }

When a request is made for this path with response validation turned on, it results in an exception in bravado_core:

    NotImplementedError: This OutgoingResponse type <class 'pyramid_swagger.tween.PyramidSwaggerResponse'> forgot to implement an attr for `headers`

This patch makes the PyramidSwaggerRequest and PyramidSwaggerResponse classes delegate all unknown attribute requests to the wrapped pyramid.Request and pyramid.Response objects, rather than just adding a `headers` property method to PyramidSwagerResponse, so that any future additional attribute reads from bravado_core can also be satisfied without further changes in this project.

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -220,10 +220,6 @@ class PyramidSwaggerRequest(RequestLike):
         return self.route_info.get('match') or {}
 
     @property
-    def headers(self):
-        return self.request.headers
-
-    @property
     def form(self):
         """
         :rtype: dict
@@ -248,6 +244,12 @@ class PyramidSwaggerRequest(RequestLike):
     def json(self, **kwargs):
         return getattr(self.request, 'json_body', {})
 
+    def __getattr__(self, name):
+        try:
+            return getattr(self.request, name)
+        except:
+            return super(PyramidSwaggerRequest, self).__getattr__(name)
+
 
 class PyramidSwaggerResponse(OutgoingResponse):
     """Adapter for a :class:`pyramid.response.Response` which exposes response
@@ -269,6 +271,12 @@ class PyramidSwaggerResponse(OutgoingResponse):
 
     def json(self, **kwargs):
         return getattr(self.response, 'json_body', {})
+
+    def __getattr__(self, name):
+        try:
+            return getattr(self.response, name)
+        except:
+            return super(PyramidSwaggerResponse, self).__getattr__(name)
 
 
 def handle_request(request, validator_map, validation_context, **kwargs):
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -247,7 +247,7 @@ class PyramidSwaggerRequest(RequestLike):
     def __getattr__(self, name):
         try:
             return getattr(self.request, name)
-        except:
+        except AttributeError:
             return super(PyramidSwaggerRequest, self).__getattr__(name)
 
 
@@ -275,7 +275,7 @@ class PyramidSwaggerResponse(OutgoingResponse):
     def __getattr__(self, name):
         try:
             return getattr(self.response, name)
-        except:
+        except AttributeError:
             return super(PyramidSwaggerResponse, self).__getattr__(name)
 
 
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -10,7 +10,7 @@ import sys
 import bravado_core
 from bravado_core.exception import SwaggerMappingError
 from bravado_core.formatter import SwaggerFormat  # noqa
-from bravado_core.request import RequestLike, unmarshal_request
+from bravado_core.request import IncomingRequest, unmarshal_request
 from bravado_core.response import get_response_spec, OutgoingResponse
 from pyramid.interfaces import IRoutesMapper
 import jsonschema.exceptions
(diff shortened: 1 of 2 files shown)


bwarfield-amplify:
Oddly, those tests pass about half the time without the most recent change, and all of the time with it, in my local build.  &lt;scratches head perplexedly&gt;

pyramid_swagger/tween.py
```diff
@@ -248,6 +244,12 @@ def files(self):
     def json(self, **kwargs):
         return getattr(self.request, 'json_body', {})
 
+    def __getattr__(self, name):
```
striglia: So if I'm reading the [Object Model](https://docs.python.org/2/reference/datamodel.html#object.__getattr__) properly, this will only get used when we are using a name which hasn't already been defined in this class? i.e. the json attribute will hit the method two lines above, not pass through this getattr func?

striglia:
@analogue I think you wrote (or helped write) a fair portion of these classes. Any context/thoughts to add from your side?

pyramid_swagger/tween.py
bwarfield-amplify: Exactly.  The test includes hitting the `body` attribute to make sure that it's still calling the right property method, not hitting the fallback.

analogue:
I can't say that I support this sort of "delegate everything that doesn't match" sort of solution.  The interfaces for `Incoming[Request|Response]` and `Outdoing[Request|Response]` are strict and precise. You're throwing caution to the wind and have two things to be concerned about:

1. You're assuming future unhandled attrs in the PyramidSwaggerResonse should automatically pass through
2. You're assuming the type of the attribute is compatible with whatever the interface declares it to be.

bwarfield-amplify:
I am making that assumption, yes—it seemed like a reasonable one from where I was sitting, but obviously that's a judgement call on the part of the maintainer, not me. :-)

I don't know the bravado-core project particularly well, so it's all guesswork from my standpoint, but I would rephrase both of those concerns to the following: "you're assuming that if the bravado-core `[Incoming|Outgoing][Request|Response]` interfaces change in the future to require additional attributes, and those attributes have the same names as attributes of the existing `webob.[Request|Response]` objects, then the type of those attributes in webob will be acceptable to bravado-core."  That does seem to me like a reasonable enough assumption to me to justify the fallback behavior, but if it still seems unreasonable and unsafe to people who know the problem space better, I'm happy to redo the PR with a more limited patch that solves the immediate problem.

bwarfield-amplify:
Removed the __getattr__ fallbacks.

Also added a bit of error message management to prevent user confusion, which is somewhat unrelated to the original purpose of the PR—can separate things out into new branches if desired.

pyramid_swagger/tween.py
```diff
@@ -264,8 +280,25 @@ def content_type(self):
...
+            return None
+        # if there's a response body, try to read it as text, but carefully
+        try:
+            return self.response.text
+        except AttributeError as prev_ex:
+            # pyramid.response.Response raises AttributeError if you try to
+            # read this property without setting the charset first: trap that
+            # so it doesn't show up as a missing "text" attribute (and get
+            # its error message swallowed by the base class)
+            raise Exception(str(prev_ex))
```
bwarfield-amplify: I wanted to re-raise this with the original stack trace, but doing that in a 2/3-compatible way requires adding `six` to the dependencies, and that didn't seem worthwhile for a single stack trace.

striglia: I have zero issue with introducing a six dep (I'm pretty committed to keeping py2/3 compat for a while), but agreed that the difference in the stack trace here isn't really interesting.

striglia:
I think this might actually be caused (or at least seemingly related!) to https://github.com/Yelp/bravado/issues/206, which @analogue mentions is a bravado-core issue. So maybe a bump of bravado-core along with this change to inherit from IncomingRequest will fix?

bwarfield-amplify:
RequestLike is currently an alias to IncomingRequest anyway—I'm not sure about the relationship to the bravado issue, though.

analogue:
@striglia Separate issue from Yelp/bravado#206 - in that case, the headers weren't being exposed at all and also needed to be transformed in the process from a Python object to a dict-like object (one of the things I pointed out in the danger of just passing things through via magic).

striglia:
Apologies for letting this one drop off my radar. 

@bwarfield-amplify: Anything remaining to be answered/done or should I take a look so we can get this into 2.3.0?

bwarfield-amplify:
I didn't have any other updates to it planned—I think I addressed @analogue's unhappiness with my original implementation, and as far as I know this works in all the environments it's supposed to work in (though the auto-test job doesn't seem to have re-run since my last update?).  So yes, as far as I know this is ready for review/merge, and if it's not OK to merge I'm happy to do whatever else is needful to get it ready in time for 2.3.0.

striglia:
Thanks for the patience Ben. I'll roll this into 2.3.0-rc2 once the other outstanding pull requests get finished up.

PR #163, opened 2016-04-08
Better messaging on route not found in get_op_for_request

atmz:
Fixes https://github.com/striglia/pyramid_swagger/issues/162 (or at least one of the cases there). Should not have any effect on performance. Also corrects docstring.

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -563,7 +563,8 @@ def get_op_for_request(request, route_info, spec):
             return op
         else:
             raise PathNotMatchedError(
-                "Could not find a matching Swagger operation for {0} request {1}"
+                "Could not find a matching Swagger "
+                "operation for {0} request {1}"
                 .format(request.method, request.url))
     else:
         raise PathNotMatchedError(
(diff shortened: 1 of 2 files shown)


djeebus:
This would have saved me a good amount of time over the last few weeks!

pyramid_swagger/tween.py
```diff
@@ -561,6 +561,11 @@ def get_op_for_request(request, route_info, spec):
         op = spec.get_op_for_request(request.method, route_path)
```
striglia: I find this logic flow slightly counter-intuitive. Can we just move the lowest-level raise up into this if block?

tests/tween_test.py
```diff
@@ -214,7 +214,7 @@ def test_get_op_for_request_not_found_when_route_has_no_path():
     swagger_spec = Mock(spec=Spec)
     with pytest.raises(PathNotMatchedError) as excinfo:
         get_op_for_request(request, route_info, swagger_spec)
-    assert 'Could not find a matching Swagger operation' in str(excinfo.value)
+    assert 'Could not find a matching route' in str(excinfo.value)
```
striglia: Maybe update the func name. Existing one is pretty mechanical (as opposed to explaining the likely programming mistake)

striglia:
🎉

Issue #147, opened 2015-11-11
Support multiple json files for Swagger 2.0


analogue:
Currently, we only support Swagger 2.0 specs that are wholly contained in a single ``swagger.json`` file.  As part of [bravado-core's refactoring to support recursive $refs] (https://github.com/Yelp/bravado-core/issues/35), serving up a swagger.json that references other json files also becomes possible. As such, clients will create http requests for other files that embedded $refs point to.

To support this feature, pyramid-swagger needs to be able to serve up all files that make up a spec and register the appropriate views.  See [register_api_doc_endpoints] (https://github.com/striglia/pyramid_swagger/blob/ae2de7770d2eced733490adbdf51cf1d895f92af/pyramid_swagger/__init__.py#L46)

striglia:
Thanks Semir. Can you briefly describe the necessary change here? Presumably at least a bump to the bravado version, although we don't explicitly limit it in this library. Perhaps some documentation?

analogue:
I've prototyped the change in this [branch](https://github.com/striglia/pyramid_swagger/compare/master...analogue:recursive_refs?expand=1) to verify the version of bravado-core (4.0.0) that supports recursive refs works as expected. An end-to-end test using the internal example_happyhour was successful. 

Since pyramid view callables can't be ``partials`` I used a rather ugly hack to deduce the json file name from the request's path.

Issue #160, opened 2016-03-31
Recursive object definitions in spec cause a stack overflow in view_for_swagger_schema (2.3.0rc1)


bwarfield-amplify:
Using the schema below, if I try to view /swagger.json, I get a RuntimeError from excessive recursion.  This does not occur in 2.2.3.  Last few lines of the stack trace:

      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 113, in resolve_refs
        return resolve_ref(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 104, in resolve_ref
        return resolve_refs(spec, spec_dict)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 113, in resolve_refs
        return resolve_ref(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 104, in resolve_ref
        return resolve_refs(spec, spec_dict)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
      File "/Users/bwarfield/.virtualenvs/popeye/lib/python2.7/site-packages/pyramid_swagger/api.py", line 115, in resolve_refs
        new_dict[key] = resolve_refs(spec, subval)
[...]

PR #164, opened 2016-04-09
Serve full schemas, individually

djeebus:
# High level overview
This adds support for serving up all the external files that make up a swagger schema, and the ability to translate between json and yaml transparently.

In other words, if you have a `swagger.yaml` file that references a `models.json` file, all of the following requests return files:
- `GET /swagger.yaml` (the external reference will be rewritten to `{$ref: models.yaml}` to match the requested format
- `GET /swagger.json` (with the `$ref` rewritten to `models.json`)
- `GET /models.json`
- `GET /models.yaml`

This also inherently solves the recursive schema issue, #160 

# Low level overview
This was a little tougher to accomplish than repackaging the entire schema as a single file. The changes can be broken down into three chunks: 
a) creating all the appropriate routes, 
b) serving up all the schemas
c) hiding the api doc routes from the tween

## Routes
This requires walking the `swagger.json` for all `$ref` nodes, and identifying which urls should be served by us. I decided that only relative $refs should be served (if a $ref is hosted on a different website, we can't rewrite the contents of it easily). `NodeWalkerForRefFiles` is the class that implements finding all the relative file refs. 

## Serving up the schemas
This requires rewriting the `$refs` before returning the dict, in order to keep the client consistent. A client requesting a `swagger.json` file would be very confused if it received a `models.yaml` file, I think. The class responsible for walking the schema and rewriting the `$refs` is `NodeWalkerForCleaningRefs`. 

## Hiding the API doc route from the tween
This was implemented in `is_swagger_documentation_route`. All the api doc routes have the same prefix; this finds those and skips schema validation for those.

### DIFF
--- a/tests/acceptance/api_test.py
+++ b/tests/acceptance/api_test.py
@@ -95,3 +95,4 @@ def test_recursive_swagger_api_external_refs():
     }))
 
     recursive_test_app.get('/swagger.json', status=200)
+    recursive_test_app.get('/external.json', status=200)
--- a/tests/acceptance/api_test.py
+++ b/tests/acceptance/api_test.py
@@ -77,3 +77,21 @@ def test_default_only_serves_up_swagger_20_schema(default_test_app):
     # swagger 1.2 schemas should 404
     for path in ('/api-docs', '/api-docs/sample', '/api-docs/other_sample'):
         default_test_app.get(path, status=404)
+
+
+def test_recursive_swagger_api_internal_refs():
+    recursive_test_app = TestApp(main({}, **{
+        'pyramid_swagger.schema_directory':
+            'tests/sample_schemas/recursive_app/internal/',
+    }))
+
+    recursive_test_app.get('/swagger.json', status=200)
+
+
+def test_recursive_swagger_api_external_refs():
+    recursive_test_app = TestApp(main({}, **{
+        'pyramid_swagger.schema_directory':
+            'tests/sample_schemas/recursive_app/external/',
+    }))
+
+    recursive_test_app.get('/swagger.json', status=200)
--- a/tests/acceptance/relative_ref_test.py
+++ b/tests/acceptance/relative_ref_test.py
@@ -1,7 +1,11 @@
 # -*- coding: utf-8 -*-
+import json
+import os.path
 import pytest
-from webtest import TestApp
+import yaml
 
+from six import BytesIO
+from webtest import TestApp
 from .app import main
 
 
@@ -25,3 +29,55 @@ def test_app(settings):
 def test_running_query_for_relative_ref(test_app):
     response = test_app.get('/sample/path_arg1/resource', params={},)
     assert response.status_code == 200
+
+
+def fix_ref(ref, schema_format):
+    if schema_format == 'json':
+        return ref  # all refs are already yaml
+    return ref.replace('.json', '.%s' % schema_format)
+
+
+def walk_schema_for_refs(schema_item, schema_format):
+    if isinstance(schema_item, dict):
+        for key, value in schema_item.items():
+            if key == '$ref':
+                schema_item[key] = fix_ref(value, schema_format)
+            else:
+                walk_schema_for_refs(value, schema_format)
+    elif isinstance(schema_item, list):
+        for item in schema_item:
+            walk_schema_for_refs(item, schema_format)
+
+
+@pytest.mark.parametrize('schema_format', ['json', 'yaml'])
+def test_swagger_schema_retrieval(schema_format, test_app):
+    here = os.path.dirname(__file__)
+    deserializers = {
+        'json': lambda r: json.loads(r.body.decode('utf-8')),
+        'yaml': lambda r: yaml.load(BytesIO(r.body)),
+    }
+
+    deserializer = deserializers[schema_format]
+
+    expected_files = [
+        'parameters/common',
+        'paths/common',
+        'responses/common',
+        'swagger',
+    ]
+    for expected_file in expected_files:
+        response = test_app.get('/%s.%s' % (expected_file, schema_format))
+        assert response.status_code == 200
+
+        gold_path_parts = [
+            here, '..', 'sample_schemas', 'relative_ref',
+            '%s.json' % expected_file,
+        ]
+        with open(os.path.join(*gold_path_parts)) as f:
+            expected_dict = json.load(f)
+
+        walk_schema_for_refs(expected_dict, schema_format)
+
+        actual_dict = deserializer(response)
+
+        assert actual_dict == expected_dict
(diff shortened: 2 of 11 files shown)


striglia:
Definitely a lot of code! Thanks again for putting it together.

Can you give a brief (maybe 3-4 sentence) explanation of how this code is structured and how it accomplishes the goal? I think that context would help me get a better mental model and hopefully get this out the door quicker :)

djeebus:
@striglia sorry about the anemic documentation; added a bunch of description to the PR.

striglia:
No worries, this is a pretty major feature and I want to make sure I follow :) Thanks for the extra work

tests/tween_test.py
```diff
@@ -337,3 +337,8 @@ def test_get_swagger12_objects_if_both_present_but_route_not_in_prefer20(
     swagger_handler, spec = get_swagger_objects(settings, route_info, registry)
     assert 'swagger12_handler' in str(swagger_handler)
     assert 'schema12' == spec
+
+
+def test_is_swagger_documentation_route_without_route_is_safe():
```
striglia: Mind a brief docstring on why this is worth testing? (e.g. what real-world case it applies to)

tests/includeme_test.py
```diff
@@ -43,6 +43,7 @@ def test_bad_schema_not_validated_if_spec_validation_is_disabled(_):
     settings = {
         'pyramid_swagger.schema_directory': 'tests/sample_schemas/bad_app/',
         'pyramid_swagger.enable_swagger_spec_validation': False,
+        'pyramid_swagger.enable_api_doc_views': False,
```
striglia: How come?

tests/api_test.py
```diff
@@ -50,3 +50,18 @@ def test_proper_error_on_missing_api_declaration():
         'tests/sample_schemas/missing_api_declaration/missing.json'
         in str(exc)
     )
+
+
+def test_ignore_absolute_paths():
```
striglia: Let's stick a comment to this PR either in this test or in the code it covers so our future selves have some clue why we have this behavior :)

tests/acceptance/yaml_test.py
```diff
@@ -49,118 +51,57 @@ def test_user_format_failure_case(testapp_with_base64):
...
+    assert yaml.load(response.body) == expected_dict
+
+
+def fix_ref(ref, schema_format):
+    if schema_format == 'yaml':
+        return ref  # all refs are already yaml
+    return ref.replace('.yaml', '.%s' % schema_format)
+
+
+def walk_schema_for_refs(schema_item, schema_format):
```
striglia: "walk_schema_for_refs" confused me...maybe "recursively_fix_refs"? Also is this just a helper to take a schema and make it look like it has external YAML/JSON refs easily? Might be worth a brief explanation in docstring or the consuming test

tests/acceptance/relative_ref_test.py
```diff
@@ -25,3 +29,55 @@ def test_app(settings):
...
+            '%s.json' % expected_file,
+        ]
+        with open(os.path.join(*gold_path_parts)) as f:
+            expected_dict = json.load(f)
+
+        walk_schema_for_refs(expected_dict, schema_format)
+
+        actual_dict = deserializer(response)
+
+        assert actual_dict == expected_dict
```
striglia: Will it be self-evident which format was to blame if this assert fails? I forget how obviously parameterized test data is included in errors...

tests/acceptance/api_test.py
```diff
@@ -77,3 +77,21 @@ def test_default_only_serves_up_swagger_20_schema(default_test_app):
...
+    recursive_test_app.get('/swagger.json', status=200)
+
+
+def test_recursive_swagger_api_external_refs():
+    recursive_test_app = TestApp(main({}, **{
+        'pyramid_swagger.schema_directory':
+            'tests/sample_schemas/recursive_app/external/',
+    }))
+
+    recursive_test_app.get('/swagger.json', status=200)
```
striglia: Do we need to add tests that we can access multi-endpoint docs like /models/pets.json and /models/pets.yaml?

djeebus:
Thanks for the code review! I'll get on these soon.

tests/tween_test.py
djeebus: This one is mostly my uncertainty as to what comes out of `IRoutesMapper`; ie if it can ever be `None` or if it's always an empty `dict`. If it's always an empty `dict`, I can remove this test.

tests/includeme_test.py
djeebus: `pyramid_swagger.get_swagger_spec` was `mock`'d out in order to avoid testing its details with a bad spec. If you enable the API doc views, it attempts to parse the posix path of spec refs, which ends up trying to compare two `MagicMock` objects. This succeeds in py27, but fails in py35. I assumed this test wasn't trying to catch those types of errors, so I just told it to ignore the api doc routes. Lots of other tests verify those properly anyway.

tests/acceptance/relative_ref_test.py
djeebus: Yeah, real obvious. The following is the result of changing 'yaml' to 'yamlses' and running that test. You'll see the format in the test header.

```
________________________________________________ test_swagger_schema_retrieval[yamlses] _________________________________________________

schema_format = 'yamlses', test_app = <webtest.app.TestApp object at 0x7faee64ebb10>

    @pytest.mark.parametrize('schema_format', ['json', 'yamlses', ])
    def test_swagger_schema_retrieval(schema_format, test_app):
        here = os.path.dirname(__file__)
        deserializers = {
            'json': lambda r: json.loads(r.body.decode('utf-8')),
            'yaml': lambda r: yaml.load(BytesIO(r.body)),
        }
    
>       deserializer = deserializers[schema_format]
E       KeyError: 'yamlses'

tests/acceptance/relative_ref_test.py:64: KeyError
```

tests/acceptance/api_test.py
djeebus: The test `test_swagger_schema_retrieval` runs through 4 files and makes sure all are accessible, but just for good measure I added a second call to `test_recursive_swagger_api_external_refs`.

tests/api_test.py
djeebus: done

djeebus:
Not sure what's going on w/ the docs build, but everything else passes :(

tests/includeme_test.py
striglia: Gotcha. Sounds good

tests/acceptance/api_test.py
striglia: 👍

striglia:
Ah docs are my fault...that's what I get for trying to add Github links apparently :)

I'll fix those up and (fingers crossed) we can release rc2 and test it out!

striglia:
Closes #147

striglia:
@djeebus @bwarfield-amplify Let me know if you have any feedback on rc2 (positive or negative)! Thanks for making it possible :)

https://pypi.python.org/pypi/pyramid_swagger/

PR #169, opened 2016-06-22
Add serve_dereferenced_swagger_schema parameter

ATRAN2:
Add an option to serve full swagger schemas with all schema refs
resolved (The behavior prior to pyramid_swagger 2.3.0-rc2).

During a deployment where servers are distributed, there can be a transient period where some servers have an old version of a schema and a new version of a schema.  In pyramid_swagger 2.3.0-rc2, a schema is built from multiple requests from each ref.  If a new schema tries to resolve a ref but points to an old schema (or vice versa) during the transient deploy period, there can be issues in the resulting schema.  This pull request adds an option to use the old behavior prior to 2.3.0-rc2 where the server resolves all the refs and serves the fully dereferenced schema upon a request for swagger.{json, yaml}

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -71,6 +71,12 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # below for more details
         pyramid_swagger.generate_resource_listing = false
 
+        # Enable/disable serving the entire dereferenced swagger schema in
+        # a single http call. This can be slow for larger schemas and will
+        # not work for infinitely recursive refs
+        # Default: False
+        pyramid_swagger.dereference_served_schema = False
+
 
 Note that, equivalently, you can add these settings during webapp configuration:
 
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -101,6 +101,29 @@ def build_swagger_12_api_declaration_view(api_declaration_json):
     return view_for_api_declaration
 
 
+def resolve_ref(spec, url):
+    with spec.resolver.resolving(url) as resolved_spec_dict:
+        spec_dict = copy.deepcopy(resolved_spec_dict)
+        return resolve_refs(spec, spec_dict)
+
+
+def resolve_refs(spec, val):
+    if isinstance(val, dict):
+        new_dict = {}
+        for key, subval in val.items():
+            if key == '$ref':
+                # assume $ref is the only key in the dict
+                return resolve_ref(spec, subval)
+            else:
+                new_dict[key] = resolve_refs(spec, subval)
+        return new_dict
+
+    if isinstance(val, list):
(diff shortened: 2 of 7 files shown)


analogue:
This is a sticky problem to solve in a sustainable way. I wouldn't want to rely on the old behaviour unless you have no other option. Note that this will blow up on you the moment your schema contains a recursive ref. Also, I'd verify all the x-model stuff works without issue in the legacy behaviour mode.

striglia:
I'm game for the flag...feels like it rounds out 2.3.0 with the necessary features for yaml/multi-file support to be tenable. A couple comments inline, but let's include it (with sufficient warnings about where it's a bad idea).

tests/acceptance/relative_ref_test.py
```diff
@@ -87,3 +101,26 @@ def test_swagger_schema_retrieval(schema_format, test_app):
         actual_dict = deserializer(response)
 
         assert actual_dict == expected_dict
+
+
+@pytest.mark.parametrize('schema_format', ['json', 'yaml', ])
+def test_dereferenced_swagger_schema_retrieval(schema_format, test_app_deref):
```
striglia: Have you confirmed this fails without the test_app_deref?

pyramid_swagger/api.py
```diff
@@ -233,6 +256,37 @@ def __call__(self, value, system):
 
 
 def build_swagger_20_swagger_schema_views(config):
+    settings = config.registry.settings
+    if settings.get('pyramid_swagger.serve_dereferenced_swagger_schema'):
```
striglia: Maybe just "dereference_served_schema" for the config name?

pyramid_swagger/api.py
```diff
@@ -101,6 +101,29 @@ def view_for_api_declaration(request):
     return view_for_api_declaration
 
 
+def resolve_ref(spec, url):
```
striglia: Can we adjust the configuration.rst page accordingly to make sure we well-explain when and why to use this?

ATRAN2: Included an entry for pyramid_swagger.dereference_served_schema in configuration.rst

docs/configuration.rst
```diff
@@ -71,6 +71,12 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # below for more details
         pyramid_swagger.generate_resource_listing = false
 
+        # Enable/disable serving the entire dereferenced swagger schema in
+        # a single http call. This can be slow for larger schemas and will
+        # not work for infinitely recursive refs
```
striglia: Link to the relevant issue describing recursive refs?

ATRAN2: This is for refs that would have a ref property to itself.  This would lead to an infinite recursion and error out

PR #170, opened 2016-06-30
Allow recursive references

macisamuele:
With the last release was added the ``dereference_served_schema`` parameter.
As described by the documentation the parameter is not working well in case of recursive model definition.

*The pull request is trying to fix this issue.*

The idea is to create a big self-containing swagger specification which could contain ``$ref`` tags.
Having internal references allows the generation of recursive models and without preventing the generation of a unique swagger specification file

Ping: @striglia, @lucagiovagnoli, @sjaensch

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -105,7 +105,8 @@ def build_swagger_12_api_declaration_view(api_declaration_json):
 
 def dereference_definition(spec, url, current_file, definitions_dict):
     """
-    Dereference a given definition (by URL) updating the definitions dictionary and fetching all the related objects
+    Dereference a given definition (by URL) updating the definitions
+    dictionary and fetching all the related objects
     :param spec:
     :param url: url of the reference to be extracted
     :param current_file: base file used for relative url resolution
@@ -115,7 +116,8 @@ def dereference_definition(spec, url, current_file, definitions_dict):
 
     def _translate_definition_target(target, prefix='', separator='|'):
         """
-        Translate the definition target to a consistent way which not uses the '#' for the name definition
+        Translate the definition target to a consistent way
+        which does not uses the '#' for the name definition
         :param target: original definition target
         :param prefix: prefix to be used if no file were specified on the target
         :param separator: separator character to be used instead of '#'
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -269,8 +269,7 @@ def _unmarshal_target(marshaled_target):
         )
 
 
-def _resolve_references_rec(spec, base_json, file_path,
-                            defs_dict, json_path):
+def _resolve_references_rec(spec, base_json, file_path, defs_dict):
     """
     Get self-contained swagger specifications of the base_json dictionary.
     The resulting swagger specifications are equivalent to the original specs,
@@ -284,8 +283,6 @@ def _resolve_references_rec(spec, base_json, file_path,
     :type file_path: str
     :param defs_dict: known definitions
     :type defs_dict: dict
-    :param json_path: JSON structure path of base_json
-    :type json_path: str
     :return: swagger specification targeting definitions in defs_dict
     """
 
@@ -315,12 +312,11 @@ def _resolve_references_rec(spec, base_json, file_path,
                         _extract_reference(target),
                         target_url.path,
                         defs_dict,
-                        '{target}/'.format(target=target_url.fragment),
                     )
                     defs_dict[target_name] = resolved
 
                 result_dict[key] = '#/definitions/{target}'.format(
-                    target=_marshal_target(target_url),
+                    target=target_name,
                 )
             else:
                 result_dict[key] = _resolve_references_rec(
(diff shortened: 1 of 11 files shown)


pyramid_swagger/api.py
```diff
@@ -101,29 +103,104 @@ def view_for_api_declaration(request):
...
+        """
+        Translate the definition target to a consistent way
+        which does not uses the '#' for the name definition
+        :param target: original definition target
+        :param prefix: prefix used if no file were specified on the target
+        :param separator: separator character to be used instead of '#'
+        :return: the converted target (it will be the new name of the link)
+        """
+        value = target.replace("#/definitions/", separator)
+        if value.split(separator)[0] == '':
```
sjaensch: Isn't this the same as `if value[0] == separator:` as long as separator is a char?

macisamuele: Sure ... 😄

sjaensch:
I really like what this PR is doing! However I'm wondering about test coverage. What tests cover this behavior? I'm specifically looking for one where we'd have a multi-file spec, it gets ingested by pyramid_swagger and we assert on the resulting spec as it would have been served by pyramid_swagger.

striglia:
+1 to Stephan's goal of ensuring we have new tests that are failing without your code change and passing with it. This was mostly a known issue w/ the previous behavior, but we can at least ensure it doesn't creep back in :)

sjaensch:
One other potential concern: did you test this with a client, e.g. bravado? Not sure if the slashes you use within the new reference name might be interpreted as a path during resolution.

macisamuele:
So, I have just pushed some modification on the code in order to:
 - have a smarter generation of names by ``_translate_definition_target`` (it is highly customizable and replaces ``/`` with ``.``) and takes care about the relative paths respect to ``swagger.json`` (have a cleaner naming process)
 - make more evident the dereferencing process through the test files on ``/tests/sample_schemas/relative_ref/``

About the bravado client testing I have to test it with the new version of the code.

macisamuele:
I'm sorry for the code modifications that were introduced by the last commits!
Them were necessary in order to produce a more readable and stable code and to reduce the amount of information that have to be propagated by the recursive calls.

At the current state the generated swagger specs (due to the ``dereference_served_schema`` parameter) are self-contained.
Be self-contained means that the ``$ref`` tags have to target the current file.
Since we can still use the ``$ref``s the code is fetching all the *external* references and putting them on the ``/definition/`` part of the specs. 

An additional test with the bravado client is ensuring that the generated swagger spec are compliant.

macisamuele:
Additional tests on commit c50aeda have highlighted an additional issue.

The issue was related to the fact that everything is inserted in the ``definitions`` JSON object, which lead to a correct JSON representation which unfortunately is not respecting all the Swagger Specifications.

For example the following JSON is not valid Swagger:
```JSON
{
  "info": {
    "version": "1.0.0",
    "title": "title"
  },
  "paths": {
    "/test1/": {
      "$ref": "#/definitions/file.paths.json|..paths..test1"
    }
  },
  "schemes": [
    "http"
  ],
  "produces": [
    "application/json"
  ],
  "basePath": "/",
  "definitions": {
    "file.paths.json|..paths..test1": {
      "get": {
        "operationId": "test1",
        "responses": {
          "200": {
            "description": "Return integer",
            "schema": {
              "type": "integer"
            }
          }
        }
      }
    }
  },
  "swagger": "2.0"
}
```

In order to avoid it we need to dereference the [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject)s and to insert into the definitions the [Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject)s and [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#referenceObject)s. The discrimination is provided by the ``_is_model_definition`` function.

tests/sample_schemas/relative_ref/dereferenced_swagger.json
```diff
@@ -1,77 +1,98 @@
...
-                                "raw_response": {
-                                    "type": "string"
-                                },
-                                "logging_info": {
-                                    "type": "object"
-                                }
-                            }
-                        }
-                    }
+  "swagger": "2.0",
```
sjaensch: Is there a reason why you changed the indendation of this file?

tests/sample_schemas/relative_ref/dereferenced_swagger.json
```diff
@@ -1,77 +1,98 @@
...
+            "description": "Return a standard_response"
+          }
+        },
+        "operationId": "no_models_get",
+        "description": ""
+      }
+    }
+  },
+  "definitions": {
+    "file.paths..common.json|..definitions..path_args": {
```
sjaensch: I'm not sure what you're doing conforms to the Swagger spec: What you're describing here is a Parameter object, while according to the spec `definitions` needs to be a name -> Schema object mapping. This might cause issues with tools that are not as lenient as bravado_core / swagger_schema_validator.

tests/sample_schemas/relative_ref/dereferenced_swagger.json
macisamuele: The file is not only re-indented, since its content is changed and adapted to the new output format produced by the ``resolve_references`` function.

macisamuele:
The code change is too extensive and could lead to inconsistent specs (even starting from consistent specs).
The [PR 171](https://github.com/striglia/pyramid_swagger/pull/171) is intended to solve the same issue in a cleaner and safer fashion.

PR #171, opened 2016-07-08
Allow recursive references in single file specs aggregation.

macisamuele:
With the last release was added the ``dereference_served_schema`` parameter.
As described by the documentation the parameter is not working well in case of recursive model definition.

The PR includes the possibility to generate a single file swagger specs even in case of recursive model definition. It reaches the goal allowing local references only for the [definition objects](http://swagger.io/specification/#definitionsObject).

The generated specs are tested respect to the [bravado](https://github.com/Yelp/bravado) client.

Ping: @striglia, @lucagiovagnoli, @sjaensch

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -101,26 +101,206 @@ def build_swagger_12_api_declaration_view(api_declaration_json):
     return view_for_api_declaration
 
 
-def resolve_ref(spec, url):
+def _low_level_translate(path, is_marshaling=True):
+    # replacement of the scheme (valid only in un-marshaling phase)
+    scheme_replacements = [  # defined as tuples (un-marshaled, marshaled)
+        ('file://', 'file.'),
+        ('http://', 'http.'),
+        ('https://', 'https.'),
+    ]
+    # replacement over the whole string
+    replacement_patterns = [  # defined as tuples (un-marshaled, marshaled)
+        ('/', '..'),
+        ('#', '|'),
+    ]
+
+    # repl_idx allow to prevent if statements over the code execution
+    # instead of doing:
+    #       if is_marshaling:
+    #           x = x.replace(replacement[0], replacement[1])
+    #       else:
+    #           x = x.replace(replacement[1], replacement[0])
+    # we can achieve the same doing
+    #       x = x.replace(replacement[repl_idx], replacement[1-repl_idx])
+    repl_idx = 0 if is_marshaling else 1
+    replaced_scheme = ''
+    for replacement in scheme_replacements:
+        if path.startswith(replacement[repl_idx]):
+            replaced_scheme = replacement[1 - repl_idx]
+            path = path[len(replacement[repl_idx]):]
+            break
+    for replacement in replacement_patterns:
+        path = path.replace(replacement[repl_idx], replacement[1 - repl_idx])
+    return replaced_scheme + path
+
+
+def _get_absolute_link(spec, relative_path, current_path=''):
+    spec_url = urlparse(spec.origin_url)
+    spec_file = os.path.abspath(spec_url.path)
+    spec_dir = os.path.dirname(spec_file)
+
+    current_path_url = urlparse(current_path)
+
+    target_url = urlparse(relative_path)
+    if len(target_url.scheme) > 0:
+        target_scheme = target_url.scheme
+    elif len(current_path_url.scheme) > 0:
+        target_scheme = current_path_url.scheme
+    else:
+        target_scheme = 'file'
+
+    if target_scheme == 'file':  # targeting a local file
+        # if not absolute path, make it absolute
+        if not target_url.path.startswith('/'):
+            # if path is empty, then is targeting to the current file
+            filename = target_url.path if len(target_url.path) > 0 \
+                else os.path.basename(current_path_url.path)
+            target_url = urlparse('{path}#{fragment}'.format(
+                path=os.path.join(
+                    spec_dir,
+                    os.path.dirname(current_path_url.path),
+                    filename,
+                ),
+                fragment=target_url.fragment,
+            ))
+        return target_url
+
+    elif target_scheme in ['http', 'https']:
+        remote_path = os.path.abspath(os.path.join(
+            current_path_url.path if len(current_path_url.path) > 0 else '/',
+            target_url.path
+        ))
+        return urlparse('{scheme}://{hostname}{path}#{fragment}'.format(
+            scheme=target_scheme,
+            hostname=target_url.hostname,
+            path=remote_path,
+            fragment=target_url.fragment,
+        ))
+    else:
+        return urlparse(relative_path)
+
+
+def _relpath(path, start):
+    """Return a relative version of a path
+    NOTE: Code duplicated from Python 2.7.9 implementation because the
+    default implementation available on Python 2.6.9 is bugged.
+    """
+
+    start_list = [x for x in os.path.abspath(start).split(os.path.sep) if x]
+    path_list = [x for x in os.path.abspath(path).split(os.path.sep) if x]
+
+    # Work out how much of the filepath is shared by start and path.
+    i = len(os.path.commonprefix([start_list, path_list]))
+
+    rel_list = [os.path.pardir] * (len(start_list) - i) + path_list[i:]
+
+    return os.path.join(*rel_list)
+
+
+def _get_target(spec, target, current_path=''):
+    # Note: we assume that swagger.json file is always a server-local file
+    if target == '':
+        raise ValueError('Empty target')
+
+    target_url = _get_absolute_link(spec, target, current_path)
+    target_scheme = 'file' if target_url.scheme == '' else target_url.scheme
+
+    if target_scheme == 'file':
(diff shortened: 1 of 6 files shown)


pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
...
+            path=remote_path,
+            fragment=target_url.fragment,
+        ))
+    else:
+        return urlparse(relative_path)
+
+
+def _relpath(path, start):
+    """Return a relative version of a path
+    NOTE: Code duplicated from Python 2.7.9 implementation because the
```
striglia: Interesting...can we at least link to a bug report for context?

pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
...
+    # Work out how much of the filepath is shared by start and path.
+    i = len(os.path.commonprefix([start_list, path_list]))
+
+    rel_list = [os.path.pardir] * (len(start_list) - i) + path_list[i:]
+
+    return os.path.join(*rel_list)
+
+
+def _get_target(spec, target, current_path=''):
+    # Note: we assume that swagger.json file is always a server-local file
```
striglia: If target is always a local file, is there a better name for this? I'm not sure what this function does by its current naming...perhaps a docstring would help explain too?

pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
...
+        return result
+    else:
+        raise ValueError(
+            'Invalid marshaled object: {target}'.format(
+                target=marshaled_target
+            )
+        )
+
+
+def is_a_definition(json_path):
```
striglia: What do we mean by "definition" here? If it's a swagger term, maybe we can link out to the relevant terminology in their docs? This spec unpacking logic is getting quite complicated, so I'm interested in making sure it is clear when reading through...

pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
...
+def resolve_refs(spec, val, json_path, file_path, defs_dict):
     if isinstance(val, dict):
         new_dict = {}
         for key, subval in val.items():
             if key == '$ref':
+                if is_a_definition(json_path):
+                    target_url = _get_target(spec, subval, file_path)
+                    marshaled_target = _marshal_target(target_url)
+                    if marshaled_target not in defs_dict:
+                        defs_dict[marshaled_target] = None
```
striglia: Is this setting useful? Seems like the next line should make it not matter...?

pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
     return view_for_api_declaration
 
 
-def resolve_ref(spec, url):
+def _low_level_translate(path, is_marshaling=True):
```
striglia: With a general name like this, I think a clear docstring including precisely what type of object this returns and what the function does will be extra important.

pyramid_swagger/api.py
```diff
@@ -101,26 +101,206 @@ def view_for_api_declaration(request):
...
+        raise ValueError(
+            'Invalid target: {target}'.format(target=target)
+        )
+
+
+def _marshal_target(target_url):
+    target_scheme = target_url.scheme
+    if len(target_url.path) > 0 and \
+            target_scheme in ['', 'file', 'http', 'https']:
+        value = _low_level_translate(
```
striglia: If we're just immediately returning value, I'd do it on this line instead of the temporary variable

pyramid_swagger/api.py
macisamuele: I discovered it running the tests.
Debugging the application I saw a different code between Python 2.6.9 and 2.7.9 and 3.4.2.
In the next days I'll look for more deeper information and/or the reason that lead Python to change the implementation of the function.

pyramid_swagger/api.py
macisamuele: Sure ... it was done only for debugging purposes

pyramid_swagger/api.py
macisamuele: The general idea was: we have troubles if we are dereferencing recursive structure. Since the only possible recursive structure are the data-model we can prevent the complete dereferencing for those objects.
Every reference to a data model should be dereferenced in the [swagger definitions object](http://swagger.io/specification/#definitionsObject). Reading the [Swagger specification](http://swagger.io/specification/) was possible to identify the objects which referees to a *definition* (as described in the comments).

Have you some suggestion about the level of detail that I should insert on the docstring? Could I put part of this comment into the method documentation.

pyramid_swagger/api.py
macisamuele: Maybe adding a comment here would have been beneficial 😄 
The placeholder (``None``) is needed because the ``resolve_ref`` could involve a recursive method call if there is a recursive data-model dereferencing and we need to have a stop condition for the recursion.

Since we are resolving the references

pyramid_swagger/api.py
macisamuele: I have found the [python issue #5117](http://bugs.python.org/issue5117) that seems explaining the difference in the ``os.path.relpath`` execution with Python 2.6.9, 2.7.9, 3.4.2 and 3.5.2.
A more direct approach could be to *practically* seen the problem. You can do it running the script that is shown below (are added also the results that I have on my laptop).

I have not found an *official* reply to your doubt; but I'm able to provide you a short snippet that highlights the issue.

Script
```bash
for version in 2.6.9 2.7.9 3.4.2 3.5.2; do
    pyenv global $version
    python -c "import os; import sys; print(sys.version); print(os.path.relpath('/dir1/../dir2/file.json', '/'))"
    echo "=========="
done
```

My Results
```bash
2.6.9 (unknown, Jul  1 2016, 10:58:34) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]
../dir2/file.json
==========
2.7.9 (default, Jul  1 2016, 10:52:19) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]
dir2/file.json
==========
3.4.2 (default, Jul  1 2016, 10:56:36) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]
dir2/file.json
==========
3.5.2 (default, Jul  3 2016, 11:59:24) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]
dir2/file.json
```
**NOTE**: I'm using pyenv to quickly switch between python version. Each command execution shows the used version and the result of the ``os.path.relpath('/dir1/../dir2/file.json', '/')`` command.

macisamuele:
Ping: @striglia, @lucagiovagnoli, @sjaensch
Please could you give me some feedback? In that way I can add all the required docstrings and hoping for the PR merge. Thanks a lot!

pyramid_swagger/api.py
striglia: Frankly, I'm inclined to not work around a 2.6-specific bug that only shows up for some uses of relpath. Does it break an existing test? Even if so, it might be reasonable to just omit 2.6 for that test and target deprecation of it in version 3.0.

pyramid_swagger/api.py
striglia: Cool. Just clarifying that you mean the Swagger definitions object in a docstring should be sufficient.

pyramid_swagger/api.py
```diff
@@ -101,26 +101,268 @@ def view_for_api_declaration(request):
...
+    # definition is possible if:
+    # json_path = ['definitions']
+    # json_path = [ ..., 'schema']
+    # json_path = [ ..., 'items']
+    # json_path = [ ..., 'allOf']
+    return json_path == ['definitions'] or \
+           json_path[-1] in ['schema', 'items', 'allOf']
+
+
+def resolve_ref(spec, url, json_path, file_path, defs_dict):
```
striglia: Have you done some basic speed tests before and after? The additional logic hasn't made things too much slower I hope?

pyramid_swagger/api.py
macisamuele: Cool ... tomorrow morning I'll update the doc-strings and push them.

striglia:
OK, I think the major feedback is done. I'm hoping we can resolve the `relpath` issue without re-implementing it, but otherwise the tests are our standard.

Did at least one of your tests reproduce the recursive bug **and** fail before your code changes to `resolve_ref`?

macisamuele:
To reproduce the bug you should run the ``recursive_app_test.py`` on the master branch.
To do this you can execute the following commands from the repository directory:
    - ``git checkout origin/master``
    - ``git checkout allow-recursive-references tox.ini tests/acceptance/recursive_app_test.py``

macisamuele:
@striglia: From your yesterday comments I have:
 - added the DocString for the ``is_a_definition`` method
 - renamed ``is_a_definition`` to ``is_a_swagger_definition``
 - skipped some marshaller tests if using Python 2.6
 - added a note into ``docs/configuration.rst``

striglia:
Thanks for the patience and all the code @macisamuele  :)

PR #176, opened 2016-08-05
Tox file improvement

macisamuele:
Users running tests with Python 2.6 could have problems related to twisted that ends into ``ImportError: Twisted requires Python 2.7 or later.``.

The PR is solving this issue limiting the twisted version to 15.5 in case of Python 2.6.


sjaensch:
🛳

PR #180, opened 2016-08-22
Fix pytest 3 errors and warnings, add version requirement

sjaensch:
The class FixtureRequest has moved in pytest 3, causing the tests to fail. Also there were warnings since pytest assumed TestApp was a class that contains tests.

### DIFF
--- a/tests/acceptance/api_test.py
+++ b/tests/acceptance/api_test.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import pytest
-from webtest import TestApp
+from webtest import TestApp as App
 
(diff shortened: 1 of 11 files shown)


miohtama:
I run to `TestApp` issue in other project. Just curious: do you know if there is a way to set "TestApp" string to py.test no collect list through setup.cfg or pytest.ini?

striglia:
Looks like some customization is available for your code in general via http://doc.pytest.org/en/latest/example/pythoncollection.html#change-naming-conventions. Not sure if there's a way to opt-out specific classes short of renaming them....

Issue #178, opened 2016-08-18
404s not handled correctly


tomelm:
It looks like 404s aren't being handled correct:

```
ipdb> e
<RequestValidationError at 0x2b2f5c2190a8 400 Bad Request>
ipdb> e.code
400
ipdb> e.message
'Could not find a matching route for GET request http://telmalem.dev:9200/v3/foo. Have you registed this endpoint with Pyramid?'
```

If I hit a path that doesn't exist, it raises a validation error, says it can't find the route, and then raises a 400

georgiev-martin:
That error message has a typo, too.

striglia:
Curious...I'm surprised we don't have this covered in the existing testing suite. Should be easy to extend the existing acceptance testing to cover this case, then we can dig into why the code is a generic 400 not a 404.

sjaensch:
Looks like the not yet documented `validate_path` setting could help you out: setting it to False should return a 404 for paths that are not in the spec as long as pyramid is not configured to handle them. #183 will help otherwise.

sjaensch:
This was fixed by #191.

PR #183, opened 2016-09-29
Return HTTP/404 if endpoint not defined

macisamuele:
The PR is targeting #178 .

The issue was caused by the fact that in case of path not found (`PathNotMatchedError` is raised) where re-raised a `RequestValidationError` which was inheriting from HTTPBadRequest (HTTP/400) while actually the request is targeting a resource not found.


sjaensch:
This looks sane to me. On the other hand I think we would get a very similar behavior by simply setting `validate_path` to False.

As far as I can tell the whole idea of the `validate_path` setting is to make sure we don't serve requests for paths that are not defined in the swagger spec - even if pyramid was configured to handle them. As such it would be awesome if we could add a test that makes sure this is the case.

macisamuele:
As suggested by @sjaensch I have added the test for checking the behaviour in case of ``validate_path`` is disabled.

The idea of the new test is to verify that pyramid_swagger returns the correct response for a route defined in pyramid ``/path_not_defined_by_any_swagger`` and not present in the swagger specs if ``validate_path`` is disabled.

PR #186, opened 2017-03-06
Fix error, test failures due to new webob not returning content type

### LINKED ISSUE #185: This code is causing failures for JSON responses
https://github.com/striglia/pyramid_swagger/blob/e89cf43fc91d382d9255669e0cd9f12543988c8b/pyramid_swagger/tween.py#L545-L549

Caused this ticket:

https://github.com/Pylons/pyramid/pull/2971

In WebOb 1.7 it was changed so that those content-types that don't have a charset parameter (mostly application/*) would no longer send an charset content-type parameter.

JSON is specified as always being either UTF-8 or UTF-16/32 with BOM. There is no charset parameter, and clients are explicitly told to ignore it. To be a good internet citizen you should be strict in what you send... and WebOb 1.7 really tried to make that a reality.

WebOb 1.7 introduced `Response.default_body_encoding` so that `Response.text` may be used even in cases where there is no charset defined for the content-type, although I wouldn't necessarily encourage this use, instead I would recommend setting response.body with a binary type string instead after properly encoding the data.

Thanks!

sjaensch:
webob >= 1.7 does not return a content type, while we require one during validation. This makes sure pyramid_swagger as well as our tests run with both an older and a new version of webob. I ran the tests with the latest versions of webob (1.7.1) and pyramid (1.8.2), webob 1.6.3 and pyramid 1.7.1 as well as the latest version of webob and pyramid 1.7.1. pyramid 1.8 requires webob 1.7.

Fixes #185

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -543,9 +543,9 @@ def validate_response(response, validator_map):
 
 def prepare_body(response):
     # content_type and charset must both be set to access response.text
-    if response.content_type is None or response.charset is None:
+    if not response.content_type:
         raise ResponseValidationError(
-            'Response validation error: Content-Type and charset must be set'
+            'Response validation error: Content-Type must be set'
         )
 
     if 'application/json' in response.content_type:
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -542,7 +542,7 @@ def validate_response(response, validator_map):
 
 
 def prepare_body(response):
-    # content_type and charset must both be set to access response.text
+    # content_type must be set to access response.text
     if not response.content_type:
         raise ResponseValidationError(
             'Response validation error: Content-Type must be set'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: python
 env:
-- TOX_ENV=py26
 - TOX_ENV=py27
 - TOX_ENV=py34
 - TOX_ENV=py27-pyramid15
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,8 @@
 Changelog
 =========
+2.3.1 (2017-XX-XX)
+++++++++++++++++++
+* Fix issue with missing content type when using webob >= 1.7 (see #185)
 
 2.3.0 (2016-09-27)
 ++++++++++++++++++
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@ Changelog
 =========
 2.3.1 (2017-XX-XX)
 ++++++++++++++++++
+* Remove support for Python 2.6, newer Pyramid versions don't support it either
 * Fix issue with missing content type when using webob >= 1.7 (see #185)
 
 2.3.0 (2016-09-27)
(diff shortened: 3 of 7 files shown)


pyramid_swagger/tween.py
```diff
@@ -543,9 +543,9 @@ def validate_response(response, validator_map):
 
 def prepare_body(response):
     # content_type and charset must both be set to access response.text
```
striglia: Let's fix the comment here

pyramid_swagger/tween.py
```diff
@@ -542,10 +542,10 @@ def validate_response(response, validator_map):
 
 
 def prepare_body(response):
-    # content_type and charset must both be set to access response.text
-    if response.content_type is None or response.charset is None:
+    # content_type must be set to access response.text
+    if not response.content_type:
```
macisamuele: @sjaensch, @striglia: what will happen if someone uses pyramid_swagger with ``webob<1.7``?
I think that it will produces errors if response.charset is not defined, so I would enforce ``webob>=1.7`` in ``setup.py``

sjaensch: What errors would you expect? I ran the test suite with `webob` 1.6.3 and it works just fine (provided you downgrade `pyramid` as well).

macisamuele: I built up a concrete example in which it could fail.
With ``pyramid<1.8`` WebOb dependency is restricted to [``WebOb>=1.3.1``](https://github.com/Pylons/pyramid/blob/1.7.4/setup.py#L43), so if we're using an _old_ pyramid version like 1.5 as defined in ``tox`` we have no dependency restriction to ``WebOb>=1.7`` so it is possible to have ``pyramid==1.5,WebOb==1.3.1`` which now will raise ``AttributeError: You cannot access Response.text unless charset is set`` if charset is not defined (if won't happen with ``WebOB>=1.7.1``).

More reference on: [pyramid_swagger/test-fail-webob](https://github.com/macisamuele/pyramid_swagger/tree/test-fail-webob) and [travis build #39](https://travis-ci.org/macisamuele/pyramid_swagger/builds/208413993).

Said that I would enforce the usage of a newer ``WebOb`` dependency. An other possibility is to enforce the use of a newer ``pyramid`` version (which pins up ``WebOb``), but I think that it would be "too much".

sjaensch: I'm a little bit confused as your travis build says it's testing Pyramid 1.7 and Webob 1.6 and fails, so I'm assuming you're not reproducing the issue you're describing...?

But anyway, if that was indeed using older versions of those libraries, that is a general incompatibility, right? Nothing that changed with this PR, as I'm not setting or unsetting charsets. And your test only fails as you mock the response object. It will never fail in "reality" as we use the Response object from `Webob`, so it will always be configured the proper way (either with or without content type).

macisamuele: 👍  perfect ... we're safe then! I wanted to make sure that we're not missing something that could break ``pyramid_swagger`` users.
**Thanks a lot for the feedback!**

PS. you're right ... In the comment I presented a possible scenario, while the test uses ``WebOb<1.7, pyramid<1.8.0`` (which is an other possible scenario 😄 )

PR #188, opened 2017-03-09
Add pre-commit, virtual environment, requirements-dev.txt

sjaensch:
This branch will add pre-commit to the pyramid_swagger development environment. I also ran the hooks on all files.

If you now type `make` it will install hooks and create a virtual environment with all necessary tools and dependencies for pyramid_swagger development. Either activate this environment with `source .activate.sh` or use [aactivator](https://github.com/Yelp/aactivator).

This enables the developer to easily run single tests or the full test suite without using tox. Furthermore, we'll now use tox that is installed in the virtualenv, not requiring it to be installed system-wide anymore.

I've also added a small Contributing section to the README.


macisamuele:
looks sane to me 🚢

striglia:
🚤

Thanks Stephan!

PR #189, opened 2017-03-09
Fix boolean config values if read from ini file

### LINKED ISSUE #182: boolean values in .ini configuration don't work -- need conversion using paste.deploy.converters
boolean settings, like this:
pyramid_swagger.enable_path_validation = False
are read as strings when configured via .ini file.
and are always True

sjaensch:
Use the pyramid `asbool` and `aslist`[functions](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/settings.html) when reading config values. This branch now includes tests to confirm that settings are read correctly as well as another fix for reading the supported swagger versions, which is read in a different function.

Fixes #182, #200

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -632,8 +632,8 @@ def get_swagger_versions(settings):
     :return: list of strings. eg ['1.2', '2.0']
     :raises: ValueError when an unsupported Swagger version is encountered.
     """
-    swagger_versions = settings.get(
-        'pyramid_swagger.swagger_versions', DEFAULT_SWAGGER_VERSIONS)
+    swagger_versions = set(aslist(settings.get(
+        'pyramid_swagger.swagger_versions', DEFAULT_SWAGGER_VERSIONS)))
 
     if len(swagger_versions) == 0:
         raise ValueError('pyramid_swagger.swagger_versions is empty')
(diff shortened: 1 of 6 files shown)


striglia:
Perhaps a touch of testing via http://stackoverflow.com/questions/32518604/gettings-settings-and-config-from-ini-file-for-pyramid-functional-testing and also manually setting the config values in a Settings object?

I'm a little paranoid of altering this before we have any test coverage. I suppose we'd already catch some of the Settings usages of these parameters because we use that in our current tests...

sjaensch:
Yeah part of it is covered by our existing tests (including acceptance tests), but I'll work on creating a test that uses an INI file.

sjaensch:
@striglia do you think this branch is good to go?

tests/acceptance/app/config.ini
```diff
@@ -0,0 +1,8 @@
+[app:main]
+use = call:tests.acceptance.app:main
+pyramid_swagger.enable_request_validation = true
+pyramid_swagger.enable_response_validation = false
+pyramid_swagger.enable_path_validation = true
+pyramid_swagger.exclude_routes = /undefined/first /undefined/second
```
striglia: Weird. But looks right by my reading of http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/environment.html

pyramid_swagger/tween.py
```diff
@@ -18,6 +18,8 @@
 from bravado_core.response import get_response_spec
```
striglia: Let's toss a new dummy changelog entry in and document this bugfix?

sjaensch: Added it. I also double-checked the issue reported in #200 and indeed we don't read that setting correctly. I also updated our documentation, as most of the examples given there are actually wrong and don't work the way they're specified.

PR #191, opened 2017-03-10
Return a 404 when a path cannot be found in the Swagger spec

### LINKED ISSUE #2: Be more precise with which 4xx error is raised
Validation currently raises HTTP 400s when 1) a route is not matched in the spec and 2) a route is matched but validation fails. This is better than nothing, but not quite semantically correct.

The missing route should definitely return a 404, and the failed validation can return either 400 or 422 (see http://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm for details).

sjaensch:
We'll only do this if path validation is turned on; otherwise Pyramid itself will return a 404 if no matching route is registered. So not only does the 404 make more sense, it also better fits with what Pyramid does.

During testing I uncovered that the `test_400_if_extra_body_args` uses the wrong URL, that test all of a sudden also started returning a 404. This is because the correct URL is `/sample`, not `/sample_post`.

This partially fixes #2.

### DIFF
--- a/pyramid_swagger/exceptions.py
+++ b/pyramid_swagger/exceptions.py
@@ -4,9 +4,14 @@ import sys
 import six
 from pyramid.httpexceptions import HTTPBadRequest
 from pyramid.httpexceptions import HTTPInternalServerError
+from pyramid.httpexceptions import HTTPNotFound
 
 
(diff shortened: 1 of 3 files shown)


macisamuele:
@sjaensch looks really good to me 🚢

PR #194, opened 2017-03-20
Use Python 3.5 for travis run

sjaensch:
I should have done this as part of the last PR, but unfortunately I noticed it too late. Python 3.5 seems to become somewhat the "standard" Python 3 version, with a lot of long-term supported distributions shipping it. Let's use it as the version we test against on travis.


macisamuele:
@sjaensch you should update also ``.travis.yml`` file to install the right python version on the test machine. An example could be found [here](https://github.com/Yelp/bravado-core/blob/master/.travis.yml#L5-L6).

PR #196, opened 2017-03-20
Fix #195 and include more information in ValidationErrors

### LINKED ISSUE #195: Swagger 2.0 validation errors don't raise RequestValidationError
Here the validation_context is inside the validation_error decorated function so any raised issues aren't caught and made generic for handling by the validation context.

```python
@validation_error(RequestValidationError)
def swaggerize_request(request, op, **kwargs):
    validation_context = kwargs['validation_context']
    with validation_context(request):
        request_data = unmarshal_request(request, op)
    return request_data
```

ClareCat:
fixed validation error issue, updated test for regression and added more information for exception handling. 

Updated acceptance test to only 206 on expected errors and the validation_context test started failing. This should help prevent regression in the future.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,11 @@
 Changelog
 =========
+2.3.2 (2017-03-27)
+++++++++++++++++++
+* Fix validation context for swagger 2.0 requests
+* Added docs for validation context
+* Original exception is accessible to validation context exceptions
+
 2.3.1 (2017-XX-XX)
 ++++++++++++++++++
 * Remove support for Python 2.6, newer Pyramid versions don't support it either
--- a/pyramid_swagger/exceptions.py
+++ b/pyramid_swagger/exceptions.py
@@ -8,15 +8,21 @@ from pyramid.httpexceptions import HTTPNotFound
 
 
 class RequestValidationError(HTTPBadRequest):
-    pass
+    def __init__(self, *args, **kwargs):
+        self.child = kwargs.pop('child', None)
+        super(RequestValidationError, self).__init__(*args, **kwargs)
 
 
 class PathNotFoundError(HTTPNotFound):
-    pass
+    def __init__(self, *args, **kwargs):
+        self.child = kwargs.pop('child', None)
+        super(PathNotFoundError, self).__init__(*args, **kwargs)
 
 
 class ResponseValidationError(HTTPInternalServerError):
-    pass
+    def __init__(self, *args, **kwargs):
+        self.child = kwargs.pop('child', None)
+        super(ResponseValidationError, self).__init__(*args, **kwargs)
 
 
 def wrap_exception(exception_class):
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -175,7 +175,7 @@ def validation_tween_factory(handler, registry):
                 request_data = swagger_handler.handle_request(
                     PyramidSwaggerRequest(request, route_info),
                     op_or_validators_map,
-                    validation_context=validation_context)
+                )
 
             def swagger_data(_):
                 return request_data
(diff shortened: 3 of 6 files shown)


pyramid_swagger/tween.py
```diff
@@ -166,15 +166,16 @@ def validator_tween(request):
...
         if settings.validate_request:
-            request_data = swagger_handler.handle_request(
-                PyramidSwaggerRequest(request, route_info),
-                op_or_validators_map,
-                validation_context=validation_context)
+            with validation_context(request, response=None):
+                request_data = swagger_handler.handle_request(
+                    PyramidSwaggerRequest(request, route_info),
+                    op_or_validators_map,
+                    validation_context=validation_context)
```
striglia: nuke this?

pyramid_swagger/exceptions.py
```diff
@@ -8,15 +8,21 @@
```
striglia: Would you add a section (or new page) to the documentation about how to use this decorator in practice? It can be pretty briefly, just giving a small paragraph on the potential reason for it and an example or two.

I'd like to make sure we clarify what exceptions you can expect and a couple example context managers.

pyramid_swagger/tween.py
```diff
@@ -309,16 +310,14 @@ def json(self, **kwargs):
...
     """Validate the request against the swagger spec and return a dict with
     all parameter values available in the request, casted to the expected
     python type.
 
     :param request: a :class:`PyramidSwaggerRequest` to validate
     :param validator_map: a :class:`pyramid_swagger.load_schema.ValidatorMap`
         used to validate the request
-    :param validation_context: a context manager for wrapping validation
-        errors
     :returns: a :class:`dict` of request data for each parameter in the swagger
```
striglia: Can you add a :raises: param here documenting that it returns the consolidated RequestValidationError?

tests/acceptance/request_test.py
```diff
@@ -29,8 +31,14 @@ def test_app(request, **overrides):
 def validation_context(request, response=None):
     try:
         yield
-    except Exception:
+    except (
```
striglia: Nice -- this will make testing these three exceptions much easier...

tests/acceptance/request_test.py
```diff
@@ -29,8 +31,14 @@ def test_app(request, **overrides):
...
+        exceptions.RequestValidationError,
+        exceptions.ResponseValidationError,
+        exceptions.PathNotFoundError,
+    ):
         raise exception_response(206)
+    except Exception:
+        raise exception_response(400)
```
striglia: For consistency, maybe assert that exc.child is not None?

tests/acceptance/request_test.py
```diff
@@ -29,8 +31,14 @@ def test_app(request, **overrides):
...
 
 
 validation_ctx_path = 'tests.acceptance.request_test.validation_context'
```
striglia: What test do we need to add so the issue you ran into is covered? (namely that one request flow should produce a RequestValidationError not a SwaggerMappignError)

ClareCat: This change covers it! There was a test before that should have been failing, but because it was catching all exceptions and not just specified ones it was passing.

striglia: ahhh that makes sense. Thumbs up on the new, more precise test contetmanager

docs/changelog.rst
```diff
@@ -1,5 +1,11 @@
 Changelog
 =========
+2.3.2 (2017-03-27)
```
striglia: Oops sorry, let's merge this w/ the 2.3.1 section below. The reason there isn't a date is we haven't actually tagged/released it yet :)

On that subject, will you want master tagged for use immediately or should we keep accumulating fixes?

ClareCat: It would be nice to tag master so we can update =)

striglia: wfm

PR #201, opened 2017-04-06
support virtual hosting

gweis:
use request.path_info for matching api paths, which allows reverse proxying an API on a subpath.

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -536,9 +536,9 @@ def _build_swagger_20_schema_views(config):
     file_map = {}
 
     def view_for_swagger_schema(request):
-        path, ext = os.path.splitext(request.path)
+        path, ext = os.path.splitext(request.path_info)
         ext = ext.lstrip('.')
-        actual_fname = file_map[request.path]
+        actual_fname = file_map[request.path_info]
         with spec.resolver.resolving(actual_fname) as spec_dict:
             clean_response = strip_xscope(spec_dict)
             ref_walker = NodeWalkerForCleaningRefs()
--- a/pyramid_swagger/load_schema.py
+++ b/pyramid_swagger/load_schema.py
@@ -229,7 +229,7 @@ class RequestMatcher(object):
         :returns: True if this matcher matches the request, False otherwise
         """
         return (
-            partial_path_match(request.path, self.path) and
+            partial_path_match(request.path_info, self.path) and
             request.method == self.method
         )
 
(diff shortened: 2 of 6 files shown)


sjaensch:
I don't fully understand how request.path and request.path_info differ - I've read the description of PATH_INFO in the CGI spec, but I'm not sure how this translates to UWSGI and Pyramid. Are you positive this change is backwards-compatible in all circumstances?

gweis:
PATH_INFO is the path within your application and SCRIPT_NAME is the path to the root of your application. Both together can be used to generate external URLs. In general SCRIPT_NAME is empty and PATH and PATH_INFO have the same value. However if you expose your application behind a reverse proxy and let's say you proxy a sub path (e.g. /myapp) to your application, then either the WSGI server or some middleware like repoze.vhm or similar can be used to set up the SCRIPT_NAME and PATH_INFO variables accordingly. E.g. a request to /mayapp/api/call should produce a SCRIPT_NAME=/myapp and PATH_INFO=/api/call (PATH would be the full thing /mapp/api/call). In this scenario request.path won't match any of the swagger api paths as it contains /myapp, but request.path_info will match, as it is relative to your app root.

How SCRIPT_NAME and PATH_INFO is generated really depends on your WSGI server and/or middle ware you use. UWSGI allows you to use mount points, there is also paste.deploy.config.PrefixMiddleware and my personally preferred way uses repoze.vhm (no configuration needed in the app deployment config)

Pyramid's request methods to generate URLs (like request.route_url etc...) are aware of these environment variables and will produce correct external URLs.

(hope that makes sense)

https://www.python.org/dev/peps/pep-0333/#environ-variables

gweis:
Also, to some degree this change allows for a workaround for https://github.com/striglia/pyramid_swagger/issues/199#issuecomment-289509493 .
You could run the API and your app as separate services and use a reverse proxy to expose the app at the domain root, and a proxy a subpath to the API. (won't solve the exact use case with including pyramid_swagger in your own project though).

sjaensch:
@striglia thoughts?

striglia:
Took me a bit, but I think I follow the setup here. I'm finding the Pyramid docs a little thin on this, but I agree that if you're aiming to make the Pyramid app clueless about the proxied nature of the subpath, this seems necessary.

@gweis how do you feel about adding a fresh acceptance test w/ a proxied app? It can be very basic, even re-using one of the existing specs and just hosting a sub-path of it using SCRIPT_NAME and PATH_INFO appropriately. That should fail nicely on the current master and pass with this diff.

gweis:
@striglia yes I thought about adding a test case like that (sorry for being lazy and not providing one)
I'll give it a go and see whether I can create one

gweis:
done

striglia:
Thanks for the patience and context @gweis 🚢

PR #205, opened 2017-05-24
fix ref resolver for nested specs and add tests

shrayus:
linked issue: https://github.com/striglia/pyramid_swagger/issues/204

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -324,6 +324,12 @@ def resolve_ref(spec, url, json_path, file_path, defs_dict):
         return resolve_refs(spec, spec_dict, json_path, file_path, defs_dict)
 
 
+def make_resolved_ref(marshaled_target):
+    return {'$ref': '#/definitions/{target}'.format(
+        target=marshaled_target,
+    )}
+
+
 def resolve_refs(spec, val, json_path, file_path, defs_dict):
     if isinstance(val, dict):
         new_dict = {}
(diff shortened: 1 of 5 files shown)


tests/api_test.py
```diff
@@ -69,3 +69,21 @@ def test_ignore_absolute_paths():
...
         '/usr/lib/shared/schema.json',
     ) is None
+
+
+def test_resolve_nested_refs():
+    """
+    Make sure we resolve nested refs gracefully and not get lost in
+    the recursion. Also make sure we don't rely on dictionary order
+    """
+    import os
```
striglia: Let's move these to the top of this python file instead of inlining them in our test

tests/api_test.py
```diff
@@ -69,3 +69,21 @@ def test_ignore_absolute_paths():
...
+    import yaml
+    from bravado_core.spec import Spec
+    from random import randint
+    from pyramid_swagger.api import resolve_refs
+    seed = randint(0, 4294967295)
```
striglia: Why random? I'm worried this will introduce a lot of randomness into our test builds (them failing some of the time but not all time)....

shrayus: So the big issue was that the test fails for some orderings of dictionaries, the random was an attempt to make sure that with the patch, no ordering of the dictionary fails, but tbh having a random seed is exactly equivalent to not setting a seed at all :P. I'll remove it.

striglia: Ideally we'd find a value for which this particular spec fails on the old code, then hardcode that :)

tests/api_test.py
```diff
@@ -69,3 +73,17 @@ def test_ignore_absolute_paths():
...
         '/usr/lib/shared/schema.json',
     ) is None
+
+
+def test_resolve_nested_refs():
+    """
+    Make sure we resolve nested refs gracefully and not get lost in
+    the recursion. Also make sure we don't rely on dictionary order
+    """
+    from pyramid_swagger.api import resolve_refs
```
striglia: Let's move this (and any others in this file) up to top level

pyramid_swagger/api.py
```diff
@@ -268,14 +268,20 @@ def _get_target_url(spec, target, current_path=''):
```
striglia: Let's add an entry to https://github.com/striglia/pyramid_swagger/blob/master/docs/changelog.rst under the "development version" describing what you fixed and a link to the issue for future readers.

I can release 2.4.0 after we merge this in.

Issue #206, opened 2017-06-05
recent patch to resolve-refs breaks resolving for specs with external references.


shrayus:
In a recent change to the way resolve-refs flattens a swagger spec, the way that the function generates the keys for the spec was broken. This caused specs that reference other files to not resolve correctly. 

The issue is that the function recursively resolves the spec and generates a key that represents the path from the current spec to the reference. This was broken in a previous patch (see issue #204) for specs with definitions in the same spec (when using the keyword "allOff"). To fix this, we rolled out version 3.0.0, but in doing so caused this recursive chain to not make the correct path to a file that is external to our current one. 

For an example of this issue:
suppose you have this spec
```
{
   "swagger": "2.0",
   "info": {
     "version": "1.0.0",
     "title": "dumb spec"
   },
   "paths": {
     "/blah": {
       "get": {
         "responses": {
           "200": {
             "description": "blah",
             "schema": {
               "$ref": "A.json"
             }
           }
         }
       }
     }
   }
 }
```
and the $ref to A.json is in another file and looks like this
```
{
   "properties": {
     "B": {
       "type": "object"
     }
   }
 }
```
when you try to call resolve_refs on this spec, you will get something like this
```
{'info': {'title': 'dumb spec', 'version': '1.0.0'},
 'paths': {'/blah': {'get': {'responses': {'200': {'description': 'blah',
                                                   'schema': {'$ref': '#/definitions/'}}}}}},
 'swagger': '2.0'}
```
take a look at "'schema': {'$ref': '#/definitions/}" and you can see that the reference is empty :(. This is no necessarily "broken" because the spec will still work and pass validation, but it is very confusing.

PR #207, opened 2017-06-06
fix external ref resolving and add tests

shrayus:
this is for issue number #206

### DIFF
--- a/pyramid_swagger/api.py
+++ b/pyramid_swagger/api.py
@@ -266,23 +266,47 @@ def _get_target_url(spec, target, current_path=''):
         )
 
 
+def _ensure_sane_key(fragment, mar_target):
+    '''
+    This function patches a nasty bug with the
+    which was corrupting out keys
+
+    explanation:
+        sometimes the fragment becomes:
+        '/definitions/:........swagger|..definitions..A'
+        and which makes the marshaled_target:
+        ':........swagger|..definitions..:........swagger|..definitions..A'
+        this is bad, it only comes about in a very specific case
+        this function esentially checks for that.
+    In the above example we want to return: ':........swagger|..definitions..A'
+
+    see tests/sample_specs/nested_defns/swagger.yaml
+    for an example spec that causes this issue.
+
+    '''
+    simple_fragment = fragment.split('/')[-1]
+    path_to_key = simple_fragment.split('..')
+    path = path_to_key[:-1]
+    key = path_to_key[-1]
+    bad_path = '..'.join(path + path + [key])
+    if bad_path == mar_target:
+        return '..'.join(path + [key])
+    return mar_target
+
+
 def _marshal_target(target_url):
     target_scheme = target_url.scheme
     target_fragment = target_url.fragment
     if len(target_url.path) > 0 and \
             target_scheme in ['', 'file', 'http', 'https']:
-        # Don't have nested refferences to the same file
-        # see tests/sample_specs/nested_defns/swagger.yaml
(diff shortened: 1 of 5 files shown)


tests/sample_schemas/external_refs/swagger.json
```diff
@@ -0,0 +1,21 @@
...
+    "title": "dumb spec"
+  },
+  "paths": {
+    "/blah": {
+      "get": {
+        "responses": {
+          "200": {
+            "description": "blah",
+            "schema": {
+              "$ref": "A.json"
```
striglia: Is this our first external ref in a test?!

tests/api_test.py
```diff
@@ -81,9 +82,47 @@ def test_resolve_nested_refs():
...
+    )
+    correct_dict = {
+        'swagger': '2.0',
+        'paths': {
+            '/blah': {
+                'get': {
+                    'responses': {
+                        '200': {
+                            'schema': {
+                                '$ref': '#/definitions/:....A.json|'
```
striglia: Hrm. Is there any way we can write this test without asserting the precise key that gets used here? Can we maybe *use* the spec dict returned by resolve_refs and try to use it to validate a request that exercises this bit of code?

As a general rule, higher level tests that assert the behavior users care about are better than lower level tests that assert implementation details...

pyramid_swagger/api.py
```diff
@@ -266,23 +266,49 @@ def _get_target_url(spec, target, current_path=''):
         )
 
 
+def _ensure_sane_key(fragment, mar_target):
+    '''
+    This function patches a nasty bug with the
+    'allOff' keyword in a spec, our resolved
```
striglia: What about this relies on allOf? Is it an issue with a number of nested keys like allOf and we're just noticing now? Does it happen with things like `additionalProperties` or `items`?

pyramid_swagger/api.py
```diff
@@ -266,23 +266,49 @@ def _get_target_url(spec, target, current_path=''):
...
+    ref keys become corrupted.
+
+    explanation:
+        sometimes the fragment becomes:
+        '/definitions/:........swagger|..definitions..A'
+        and which makes the marshaled_target:
+        ':........swagger|..definitions..:........swagger|..definitions..A'
```
striglia: Isn't it a bug that we're generating these doubly nested paths? Can we fix that directly in the target generation logic instead of patching it after the fact in this function?

tests/sample_schemas/external_refs/swagger.json
shrayus: Uhhh looks like it? :P

striglia: 😭

shrayus: jk I was wrong, there are tests for external refs! we're gucchi

tests/api_test.py
shrayus: ill just delete it, it is no longer needed

pyramid_swagger/api.py
shrayus: allOf calls _marshaled_target a bunch of times which is what causes the problem imo. I think it might also happen with additionalProperties and items, but I havent been able to reproduce it with them. Should I delete the comment?

pyramid_swagger/api.py
shrayus: I don't want to mess with the actual recursion too much because I don't know who else uses this code and I wouldn't want to change it in any way :P

PR #210, opened 2017-07-13
Fix bug where filename was not being updated when we follow a referen…

### LINKED ISSUE #208: resolve refs incorrectly resolves files with conflicting namespaces in its definitions dict
If I have this swagger spec
'''
{
    "swagger": "2.0",
    "info": {
        "title": " ",
        "version": "0.0.0"
    },
    "paths": {
        "/foo": {
            "get": {
                "responses": {
                    "200": {
                        "description": " ",
                        "schema": {
                            "$ref": "foo.json#/common"
                        }
                    }
                }
            }
        },
        "/bar": {
            "get": {
                "responses": {
                    "200": {
                        "description": " ",
                        "schema": {
                            "$ref": "bar.json#/common"
                        }
                    }
                }
            }
        }
    }
}
'''
There "common" references in two different files. Lets assume foo.json looks like this
'''
{
    "common": {
        "$ref": "#definitions/common"
    },
    "definitions": {
        "common": {
            "type": "integer"
        }
    }
}
'''
And bar.json looks like this:
'''
{
    "common": {
        "$ref": "#definitions/common"
    },
    "definitions": {
        "common": {
            "type": "string"
        }
    }
}
'''
When I call api.resolve_refs on my spec, my defs dict will look something like this:
'''
{
        ":....bar.json|..common": {
            "$ref": "#/definitions/:....bar.json|definitions..common"
        },
        ":....foo.json|..common": {
            "$ref": "#/definitions/:....foo.json|definitions..common"
        },
        ":....|definitions..common": {
            "type": "integer"
        }
    }
'''
As you can see, the definition in bar.json for common being a string has been overwritten.

shrayus:
resolve #208

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -112,3 +112,17 @@ def test_extenal_refs_no_empty_keys():
     spec = Spec.from_dict(spec_dict, path)
     flattened_spec = resolve_refs(spec, spec_dict, ['/'], 'swagger', {})
     traverse_spec(flattened_spec)
+
+
+def test_extenal_conflicting_namespaces():
+    """
+    This test ensures that we dont overwrite keys in defs_dict
+    if multiple files have an object with the same name.
+    """
+    with open('tests/sample_schemas/external_refs/conflicting_namespaces.json') as swagger_spec:
+        spec_dict = yaml.load(swagger_spec)
+    path = 'file:' + os.getcwd() + '/tests/sample_schemas/external_refs/swagger.json'
+    spec = Spec.from_dict(spec_dict, path)
+    defs_dict = {}
+    resolve_refs(spec, spec_dict, ['/'], 'swagger', defs_dict)
+    assert len(defs_dict) == 4
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,6 @@
 Changelog
 =========
-
+* Bugfix: update name of file when following a json reference in resolve_refs (see #208)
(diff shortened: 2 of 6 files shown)


tests/api_test.py
```diff
@@ -112,3 +112,18 @@ def test_extenal_refs_no_empty_keys():
     spec = Spec.from_dict(spec_dict, path)
     flattened_spec = resolve_refs(spec, spec_dict, ['/'], 'swagger', {})
     traverse_spec(flattened_spec)
+
+
+def test_extenal_conflicting_namespaces():
```
jnb: Did you check that this test fails without your fix?

tests/api_test.py
```diff
@@ -112,3 +112,18 @@ def test_extenal_refs_no_empty_keys():
...
+    """
+    This test ensures that we dont overwrite
```
jnb: Nit: I think you can probably use longer lines here?

tests/api_test.py
shrayus: yup! It fails, never been this proud of a failing test ;)

striglia:
@shrayus is this still relevant or will we fix this w/ #216?

macisamuele:
@striglia this PR is relevant to flattening code that has been removed by #214 .

striglia:
Cool @macisamuele, thought it might be outdated now. I'll close this out then, thanks!

Issue #177, opened 2016-08-18
pyramid_swagger is not returning an easily parsable exception message


tomelm:
I'm getting a `RequestValidationError` for a missing param but pyramid seems to be stringifying the error message:

```
ipdb> e
<RequestValidationError at 0x7f39c1cbeb88 400 Bad Request>
ipdb> e.__class__
<class 'pyramid_swagger.exceptions.RequestValidationError'>
ipdb> e.message
"phone is a required parameter.\n\nFailed validating 'required' in schema:\n    {'in': 'query',\n     'maxLength': 50,\n     'name': 'phone',\n     'required': True,\n     'type': 'string'}\n\nOn instance:\n    None"
```

I could do some kind of nasty string parsing to make this usable but that's not ideal. There doesn't appear to be any way to pull a well-formatted message

```
ipdb> e.args
("phone is a required parameter.\n\nFailed validating 'required' in schema:\n    {'in': 'query',\n     'maxLength': 50,\n     'name': 'phone',\n     'required': True,\n     'type': 'string'}\n\nOn instance:\n    None",)
ipdb> e.json
*** simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ipdb> e.json_body
*** simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ipdb> e.body
b''
```

tomelm:
The main reason this is an issue is because it makes it impossible for me to programmatically determine which field failed validation. In the case where we're building APIs, if I can't tell which field failed I don't have a way to structure the failure message appropriately without simply passing back the entire message body (which is not valid ideal in the case where we're going from internal => external)

gdchamal:
@tomelm did you find any solution ? I experimenting same issue when testing around parameter validation and a suitable json output structure

ergo:
Bumping this, I will probably also run into this problem soon.

ergo:
@gdchamal @tomelm, I've found that when you grab `RequestValidationError` it has `child` property that will return `ValidationError` from bravado that you can inspect for errors.

ergo:
Here is an example (maybe a bit naive) implementation that shows how to return a json response with error information:


    def exc_view(exc, request):
        error_info = exc.child
        for_json = {
            'message': getattr(error_info, 'message', u'{}'.format(error_info)),
            'validator': getattr(error_info, 'validator', None),
            'relative_schema_path': list(
[...]

PR #214, opened 2017-08-29
Use bravado-core methods for flattening specs

macisamuele:
bravado-core v4.8.0+ defined Swagger spec flattening feature ([PR #177](https://github.com/Yelp/bravado-core/pull/177)).

The core logic used by bravado-core was partially extracted from pyramid_swagger and partially improved.

The goal of this PR is to start using the flattening feature defined into bravado-core by removing a big quantity of code related to pure swagger specs handling (which should not be the main focus of this library).

Approving this change we can improve separation of concerns between:
 - ``pyramid_swagger``: handling pyramid related interactions
 - ``bravado-core``: core swagger handling

Updated tests to use ``spec.flattened_specs`` instead of ``resolve_refs``
Added new test defined into PR #209 to ensure that this PR already covers the reported bug.


pyramid_swagger/api.py
```diff
@@ -102,298 +102,6 @@ def view_for_api_declaration(request):
     return view_for_api_declaration
 
 
-def _low_level_translate(url, is_marshaling=True):
```
striglia: I very much 👍  removal of this kind of low-level detail from pyramid-swagger. Can you add an entry in the Changelog for this improvement? I usually just make a new "in development" heading to keep track of improvements since the last release.

macisamuele:
Sure thing

Il 30 ago 2017 7:48 PM, "Scott Triglia" <notifications@github.com> ha
scritto:

striglia:
Thanks Samuele!

PR #215, opened 2017-09-06
Fix tests due to bravado core flattening update

macisamuele:
This PR is needed to fix tests failure related to a bug introduced by ``bravado-core`` ([Issue #190](https://github.com/Yelp/bravado-core/issues/190)).

The test update is needed because from ``bravado-core >= 4.8.3`` responses can be collected on top level ``responses`` property.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
     include_package_data=True,
     install_requires=[
-        'bravado-core >= 4.8.0, != 4.8.3',
+        'bravado-core >= 4.8.4',
         'jsonschema',
         'pyramid',
         'simplejson',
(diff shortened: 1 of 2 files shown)


striglia:
Cool, I just set up daily re-running of the pyramid-swagger Travis job on master and sure enough it fails.

Should we just require newer than 4.8.3? I don't imagine we win much by allowing 4.8.0 - 4.8.2

macisamuele:
@striglia we could even force ``bravado-core >= 4.8.4`` but I was thinking that there are no specific reasons for preventing this versions.
If you prefer having ``>= 4.8.4`` I'll update this PR 😄

striglia:
Yeah let's just require >= 4.8.4 for simplicity 👍

striglia:
🚀 🌔

Issue #212, opened 2017-08-28
Expose helper method for extracting bravado_core Operation object from pyramid request


macisamuele:
While implementing endpoints that accept enum value in parameters could be interesting to have a framework support for extracting Operation object from pyramid request object.

I have in mind 2 main use cases for this:
1. in writing tests to ensure that enum strings (swagger specs) will be properly converted to enum python objects
2. in input handling of list of enums in case the assumption "all enum values if nothing is passed to the parameter"

I can prepare a PR for this, let me know if you have strong opinion against this.

pinging: @striglia , @sjaensch

striglia:
Can you clarify point 2 here? Point 1 is important, but can probably be accomplished without exposing these APIs formally. Are there use cases other than handling enum values?

macisamuele:
I have no other use cases in mind.

I was mainly thinking to enum use case because while writing specs often
happen that we define parameters as list of enum values (especially for
optional filters).
When the parameter is not provided the most general handling of it is to be
considered equivalent to all the valid enum values.
Without a feature like that developers need to manually keep
synchronization between specs and python.

Do you have different ideas for a similar use case?

Il 30 ago 2017 7:53 PM, "Scott Triglia" <notifications@github.com> ha
scritto:

Can you clarify point 2 here? Point 1 is important, but can probably be
accomplished without exposing these APIs formally. Are there use cases
other than handling enum values?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/striglia/pyramid_swagger/issues/212#issuecomment-326069030>,
or mute the thread

macisamuele:
@striglia I'm closing this issue because [``get_op_for_request``](https://github.com/striglia/pyramid_swagger/blob/1f5836520ebcb0f110b13b133fcee63330bc98e4/pyramid_swagger/tween.py#L591) already performs the needed operation

striglia:
Cool, happy it works for you 👍

PR #216, opened 2017-09-08
pyramid_swagger renderer

macisamuele:
While working with ``pyramid_swagger`` I noticed that the view responses are never marshaled via ``bravado-core`` methods. This means that if our specs are using ``format`` schema property the ``to_wire`` is never called (more info on [User-Defined Formats](http://bravado-core.readthedocs.io/en/latest/formats.html)).

Let's imagine that we have a view that returns a string with ``date`` format.
With the current (pre PR) pyramid_swagger the developer needs to enforce on the view the correct formatting of the object that represent the date (ie. ``datetime.date.today().isoformat()``), which could be annoying and different from what is happening with the input parameters (``request.swagger_data`` contains un-marshaled data, so the date string in the HTTP request will be converted to ``datetime.date`` object).
I already added the section in ``quickstart`` documentation page; on that page is available a longer description of this 😄 

The goal of this PR is:
 -  to define a new renderer (called ``pyramid_swagger``) that will perform view response marshaling and rendering with the default ``JSON`` renderer
 - to expose ``PyramidSwaggerRendererFactory`` class to *create* custom renderer factories which will operate on marshaled objects
 - to expose ``operation`` property in the request: gives access to ``bravado_core.operation.Operation`` in the view (#212) and allow ``PyramidSwaggerRendererFactory`` to properly marshal the response

NOTE: exposing ``operation`` property has "no overhead" because the property in injected by validator tween that already has evaluated the needed operation object (so the overhead is only caused by setting the property)

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -144,3 +144,140 @@ Otherwise, models are represented as dicts.
 
     Values in ``request.swagger_data`` are only available if
     ``pyramid_swawgger.enable_request_validation`` is enabled.
+
+Accessing Swagger Operation
+---------------------------
+
+During the implementation of an endpoint you could eventually have need of accessing the Swagger Specs that defined that specific view.
+``pyramid_swagger`` will inject in the request object a new property (that will be evaluated only if accessed) called ``operation``.
+
+``request.operation`` will be set to ``None`` for Swagger 1.2 defined endpoints, while it will be an `Operation <https://github.com/Yelp/bravado-core/blob/v4.8.4/bravado_core/operation.py#L12>`_ object if the endpoint is defined by Swagger 2.0 specs.
+
+Set pyramid view renderer
+-------------------------
+
+Using :mod:`pyramid_swagger` you will get automatic conversions between input JSON objects to easy to handle python objects.
+An example could be a swagger object property marked to use date property, the library will take care of converting the ISO 8601 date representation to an easy to handle python ``datetime.date`` object.
+
+Usually, while building the views that returns object that needs special handling (like ``datetime.date``) you are *forced* to perform manual operations to convert the python object on objects that could be serialized with the default ``pyramid`` renderers (ie. ``pyramid.`renderers.JSON``).
+Using ``pyramid_swagger.renderer.PyramidSwaggerRendererFactory`` as base renderer for your pyramid view will remove this *human intervention*, so you can focus more on defining the internal logic respect dealing with formatting.
+
+Here is an example:
+
+Let's assume that you have the following specs
+
+.. code-block:: json
+
+    {
+      "definitions": {
+        "json_object": {
+          "maxProperties": 1,
+          "minProperties": 1,
+          "properties": {
+            "date": {
+              "type": "string",
+              "format": "date"
+            },
+            "date-time": {
+              "type": "string",
+              "format": "date-time"
+            }
+          },
+          "type": "object"
+        }
+      },
+      "paths": {
+        "/echo_date": {
+          "post": {
+            "description": "Echoes the input date into response body",
+            "parameters": [
+              {
+                "format": "date",
+                "in": "formData",
+                "name": "date",
+                "type": "string"
+              }
+            ],
+            "operationId": "echo_date",
+            "responses": {
+              "200": {
+                "$ref": "#/responses/200_ok"
+              }
+            }
+          }
+        },
+        "/echo_date_time": {
+          "post": {
+            "description": "Echoes the input date-time into response body",
+            "operationId": "echo_date_time",
+            "parameters": [
+              {
+                "format": "date-time",
+                "in": "formData",
+                "name": "date_time",
+                "type": "string"
+              }
+            ],
+            "responses": {
+              "200": {
+                "$ref": "#/responses/200_ok"
+              }
+            }
+          }
+        }
+      },
+      "responses": {
+          "200_ok": {
+            "description": "HTTP/200 OK",
+            "schema": {
+            "$ref": "#/definitions/json_object"
+          }
+      },
+      "swagger": "2.0"
+    }
+
+and that your views are defined as
+
+.. code-block:: python
+
+    @view_config(route_name='echo_date', renderer='pyramid_swagger')
+    def echo_date(request):
+        input_date = request.swagger_data['date']
+        assert isinstance(input_date, datetime.date)
+        return {'date': input_date}
+
+    @view_config(route_name='echo_date_time', renderer='json')
+    def echo_date(request):
+        input_date_time = request.swagger_data['date_time']
+        assert isinstance(input_date_time, datetime.datetime)
+        return {'date-time': input_date_time}
+
+Calling the ``/echo_date`` endpoint you will receive, as expected, a response that will look like ``{"date": "2017-09-06"}``, while calling ``/echo_date_time`` endpoint you will receive an unexpected HTTP/500 as response.
+
+This happens because ``pyramid_swagger`` renderer performs the proper object unmarshalling (transforming the ``datetime.date`` object to an ISO 8601 string representation) before calling the base JSON renderer; while the default ``json`` renderer, which is not capable to transform to JSON a ``datetime.datetime`` object, will raise an exception.
+
+To workaround the limitation introduced by ``json`` formatter the developer has to *manually*  convert the ``datetime.datetime`` object to the desired string representation or adding `adapters <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#using-the-add-adapter-method-of-a-custom-json-renderer>`_ for the types not handled by ``json`` renderer.
+Either approaches are introducing an additional complexity on the service developer since he/she has to care about response formatting and could not be compliant with the service specification.
+
+In order to increase the flexibility of the new rendering feature, it is exposed ``PyramidSwaggerRendererFactory`` class which will allow you to define your own custom renderer.
+The defined renderer will operate on the marshaled, according to the Swagger Specification, response.
+
+Example of definition of a custom renderer
+
+.. code-block:: python
+
+    class MyPersonalRendererFactory(object):
+        def __init__(self, info):
+            # Initialize your factory
+            pass
+
+        def __call__(self, value, system):
+            # perform your personal rendering operations
+            # you can assume that value is a marshaled response, so already JSON serializable object
+            return rendered_value
+
+Once you have defined your own renderer you have to wrap the new renderer in ``PyramidSwaggerRendererFactory`` and register it to the pyramid framework as described by `Adding and Changing Renderers pyramid documentation <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#adding-and-changing-renderers>`_.
+
+.. code-block:: python
+
+    config.add_renderer(name='custom_renderer', factory=PyramidSwaggerRendererFactory(MyPersonalRendererFactory))
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -153,113 +153,78 @@ During the implementation of an endpoint you could eventually have need of acces
 
 ``request.operation`` will be set to ``None`` for Swagger 1.2 defined endpoints, while it will be an `Operation <https://github.com/Yelp/bravado-core/blob/v4.8.4/bravado_core/operation.py#L12>`_ object if the endpoint is defined by Swagger 2.0 specs.
 
-Set pyramid view renderer
--------------------------
+pyramid_swagger renderer
+------------------------
 
-Using :mod:`pyramid_swagger` you will get automatic conversions between input JSON objects to easy to handle python objects.
-An example could be a swagger object property marked to use date property, the library will take care of converting the ISO 8601 date representation to an easy to handle python ``datetime.date`` object.
+Using :mod:`pyramid_swagger` you will get automatic conversions of the input JSON objects to easy to handle python objects.
+An example could be a swagger object string property using the ``date`` format , the library will take care of converting the
+ISO 8601 date representation to an easy to handle python ``datetime.date`` object.
 
-Usually, while building the views that returns object that needs special handling (like ``datetime.date``) you are *forced* to perform manual operations to convert the python object on objects that could be serialized with the default ``pyramid`` renderers (ie. ``pyramid.`renderers.JSON``).
-Using ``pyramid_swagger.renderer.PyramidSwaggerRendererFactory`` as base renderer for your pyramid view will remove this *human intervention*, so you can focus more on defining the internal logic respect dealing with formatting.
+While defining the :mod:`pyramid` view that will handle the endpoint you have to make sure that the chosen renderer will be able to
+properly render your response. In the case of an endpoint that returns *objects* that requires a special handling
+(like ``datetime.date``) the developer is *forced* to:
 
-Here is an example:
+* manually convert the python object to an object that could be handled by the renderer
+* add an `adapter <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#using-the-add-adapter-method-of-a-custom-json-renderer>`_ for instructing pyramid to handle your object
+* define a custom renderer that is able to properly serialize the object
 
-Let's assume that you have the following specs
+:mod:`pyramid_swagger` provides:
+
+* a new renderer, called ``pyramid_swagger``
+* a new renderer renderer factory, called ``pyramid_swagger.renderer.PyramidSwaggerRendererFactory``
+
+----------------------------------
+How pyramid_swagger renderer works
+----------------------------------
+
+The new :mod:`pyramid_swagger` renderer is a wrapper around the default ``pyramid.renderers.JSON`` renderer.
+
+:mod:`pyramid_swagger` will receive, from your pyramid view, the object that has to be rendered, perform the marshaling operations and then call the default JSON renderer.
+
+.. note::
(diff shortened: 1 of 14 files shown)


docs/quickstart.rst
```diff
@@ -144,3 +144,140 @@ Otherwise, models are represented as dicts.
...
     Values in ``request.swagger_data`` are only available if
     ``pyramid_swawgger.enable_request_validation`` is enabled.
+
+Accessing Swagger Operation
+---------------------------
+
+During the implementation of an endpoint you could eventually have need of accessing the Swagger Specs that defined that specific view.
+``pyramid_swagger`` will inject in the request object a new property (that will be evaluated only if accessed) called ``operation``.
+
+``request.operation`` will be set to ``None`` for Swagger 1.2 defined endpoints, while it will be an `Operation <https://github.com/Yelp/bravado-core/blob/v4.8.4/bravado_core/operation.py#L12>`_ object if the endpoint is defined by Swagger 2.0 specs.
```
striglia: Oof, I need to write the necessary PR to properly remove 1.2 support :( It has become a notable burden for us. (no action needed, just noting)

docs/quickstart.rst
```diff
@@ -144,3 +144,140 @@ Otherwise, models are represented as dicts.
...
+
+This happens because ``pyramid_swagger`` renderer performs the proper object unmarshalling (transforming the ``datetime.date`` object to an ISO 8601 string representation) before calling the base JSON renderer; while the default ``json`` renderer, which is not capable to transform to JSON a ``datetime.datetime`` object, will raise an exception.
+
+To workaround the limitation introduced by ``json`` formatter the developer has to *manually*  convert the ``datetime.datetime`` object to the desired string representation or adding `adapters <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#using-the-add-adapter-method-of-a-custom-json-renderer>`_ for the types not handled by ``json`` renderer.
+Either approaches are introducing an additional complexity on the service developer since he/she has to care about response formatting and could not be compliant with the service specification.
+
+In order to increase the flexibility of the new rendering feature, it is exposed ``PyramidSwaggerRendererFactory`` class which will allow you to define your own custom renderer.
```
striglia: I think this example isn't super clear still on what my `__call__` method is operating on. Maybe a toy example so readers can see what a "marshalled response value" looks like vs what their `__call__` method might be doing?

docs/quickstart.rst
```diff
@@ -144,3 +144,140 @@ Otherwise, models are represented as dicts.
...
+During the implementation of an endpoint you could eventually have need of accessing the Swagger Specs that defined that specific view.
+``pyramid_swagger`` will inject in the request object a new property (that will be evaluated only if accessed) called ``operation``.
+
+``request.operation`` will be set to ``None`` for Swagger 1.2 defined endpoints, while it will be an `Operation <https://github.com/Yelp/bravado-core/blob/v4.8.4/bravado_core/operation.py#L12>`_ object if the endpoint is defined by Swagger 2.0 specs.
+
+Set pyramid view renderer
+-------------------------
+
+Using :mod:`pyramid_swagger` you will get automatic conversions between input JSON objects to easy to handle python objects.
+An example could be a swagger object property marked to use date property, the library will take care of converting the ISO 8601 date representation to an easy to handle python ``datetime.date`` object.
```
striglia: Hrm. Is there any way for us to avoid having to re-write every renderer to introduce the "right" pyramid-swagger behavior here? Ideally we'd want to do some sort of composition here where you combine a default pyramid renderer with pyramid-swagger marshalling. 

What if we write a single new swagger_view_config decorator that accepts a pass-through renderer option and the only thing it does is unmarshall any objects it has to_wire methods for? Then you would just rewrite all existing view_config decorators to be swagger_view_config decorators and nothing else would change?

striglia: Pyramid docs for reference https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#renderers-chapter

macisamuele: I was thinking to different approaches to achieve this result:
 - intercepting ``BeforeRender`` pyramid event ([doc](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/templates/templates.html))
 - "overriding" ``view_config`` decorator
 - modifying the ``Request``/``Response`` object propagated to views
 - defining a new Renderer (chosen approach)

The main reason why I discarded the other approaches was because this seemed to be the less invasive and back compatible one.
Here more detailed reasons:
 - ``BeforeRenderer`` event contains the value to render, but the passed variable could be any type of object returned by the view. This means that if we perform unmarshaling with the event listener then the object that will be rendered will be "modified" only in case of a dicitonary. Since the approach was not generalizable to all the possible valid responses I decided to move on. Additionally event listeners are not supposed to modify data.
 - overriding ``view_config`` will require developers to change the decoration system used in their services. Additionally ``view_config`` decorator is a delicate piece of pyramid framework and I guess that it could be changed/improved in the feature, so it could represent an overhead for the maintenance of pyramid_swagger
 - modifying the ``Request``/``Response`` seemed to be a valid solution, but the code was starting to be too complicated, so I decided to stop 😄 
 - defining a new renderer allows developers to easily plug the renderer if they want to.
    Marshaling should have a really small performance impact, but I would prefer to avoid to force all the users to use it.
    Additionally applying this new renderer will be a few line changes for the majority of the systems that I've worked on.
    In general an API service using pyramid_swagger defines a default renderer to be used if no renderer is specified in ``view_config`` decorator.

docs/quickstart.rst
macisamuele: I'll add an example with a test view that returns data and I'll add comments explaining that the content of the variables on each stage, I'll post the new iterations while the other comments are clear and addressed

macisamuele:
@striglia I have restructured a bit the documentation to make more easy to follow 😄 
Sorry for the really long delay on it.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+During the implementation of an endpoint you could eventually have need of accessing the Swagger Specs that defined that specific view.
+``pyramid_swagger`` will inject in the request object a new property (that will be evaluated only if accessed) called ``operation``.
+
+``request.operation`` will be set to ``None`` for Swagger 1.2 defined endpoints, while it will be an `Operation <https://github.com/Yelp/bravado-core/blob/v4.8.4/bravado_core/operation.py#L12>`_ object if the endpoint is defined by Swagger 2.0 specs.
+
+:mod:`pyramid_swagger` renderer
+-------------------------------
+
+Using :mod:`pyramid_swagger` you will get automatic conversions of the input JSON objects to easy to handle python objects.
+An example could be a swagger object property marked to use date property, the library will take care of converting the
```
sjaensch: ...could be a swagger object string property using the ``date`` format, ...

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+ISO 8601 date representation to an easy to handle python ``datetime.date`` object.
+
+While defining the :mod:`pyramid` view that will handle the endpoint you have to make sure that the chosen renderer will be able to
+properly render you response. In the case of an endpoint that returns *objects* that requires a special handling
```
sjaensch: you**r** response.  ...that require sepcial handling

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+(like ``datetime.date`` object) the developer is *forced* to:
```
sjaensch: I'd just remove "object" here.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+* perform manual operations to convert the python object to an object that could be handled by the renderer
```
sjaensch: manually convert the python object...

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+* add an `adapter <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#using-the-add-adapter-method-of-a-custom-json-renderer>`_ for instructing pyramid to handle your object
+* define a custom renderer that is able to deal with this serialization problem
```
sjaensch: ...is able to properly serialize the object

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+:mod:`pyramid_swagger` (starting from version X.X.X) provides:
+
+* a new renderer, called ``pyramid_swagger``
+* a new renderer factory of factories, called ``pyramid_swagger.renderer.PyramidSwaggerRendererFactory``
```
sjaensch: It doesn't really produce multiple factory classes though, right? You're just customizing an instance of your factory class. I'd just call it a new renderer factory, which makes it less confusing.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+Using this new solutions you can reduce your efforts on dealing with formatting and focus more on your business logic.
```
sjaensch: This sounds like PR/marketing to me, you already explained what it does.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+The new :mod:`pyramid_swagger` renderer is a wrapper around the default ``pyramid.renderers.JSON`` renderer.
+
+:mod:`pyramid_swagger` will receive, from your pyramid view, the object that has to be rendered, perform the marshaling operations and then call the default JSON renderer.
+
+.. note::
+    The usage of this renderer will allow you to better use the Custom Formats configurable with ``pyramid_swagger.user_formats``.
+
+
+Let' assume that your view returns ``{'date': datetime.date.today()}`` and that your response spec is similar to
```
sjaensch: Let's

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+-----------------------------------------
+How :mod:`pyramid_swagger` renderer works
+-----------------------------------------
+
+The new :mod:`pyramid_swagger` renderer is a wrapper around the default ``pyramid.renderers.JSON`` renderer.
+
+:mod:`pyramid_swagger` will receive, from your pyramid view, the object that has to be rendered, perform the marshaling operations and then call the default JSON renderer.
+
+.. note::
+    The usage of this renderer will allow you to better use the Custom Formats configurable with ``pyramid_swagger.user_formats``.
```
sjaensch: How? I'd add an explanation as well if you mention it. Otherwise you could say that only the pyramid_swagger renderer provides full support for custom formats...?

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+If your view is configured to use ``pyramid_swagger`` renderer then your endpoint will provide HTTP/200 responses similar
+to ``{"date": "2017-09-16"}``.
+
+This is possible because the marsharling of the view return value converts the ``datetime.date`` object to its ISO 8601
+string representation that could be handled by the default JSON renderer.
+
+.. note::
+
+    The marshaling operation will be performed according to the specific response schema defined for the particular endpoint.
+    It means that if your response doesn't specify a field it will transparently moved to the wrapped renderer.
```
sjaensch: it will be transparently passed to the wrapped renderer.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+----------------------------------------------
+How :mod:`PyramidSwaggerRendererFactory` works
+----------------------------------------------
+
+To increase the flexibility of the new rendering feature, it is exposed ``PyramidSwaggerRendererFactory`` class which will allow you to
+connect your custom defined renderer after the marshaling operation of the view result.
```
sjaensch: PyramidSwaggerRendererFactory allows you to create a custom renderer that operates on the marshaled result from your view.

docs/quickstart.rst
```diff
@@ -144,3 +144,111 @@ Otherwise, models are represented as dicts.
...
+
+The defined renderer will operate on the marshaled, according to the Swagger Specification, response.
+
+Example of definition of a custom renderer
+
+.. code-block:: python
+
+    class MyPersonalRendererFactory(object):
```
striglia: I'm not sure this is a factory at all...just looks like a renderer to me?

docs/quickstart.rst
striglia: +1

pyramid_swagger/renderer.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+            return marshal_schema_object(
+                swagger_spec=request.registry.settings['pyramid_swagger.schema20'],
+                schema_object_spec=response_spec['schema'],
+                value=response_object,
+            )
+        except MatchingResponseNotFound:
+            # Response specs not found
+            return response_object
+        except SwaggerMappingError:
```
striglia: Under what circumstances will this occur? Is silently no-oping the right response?

tests/acceptance/request_test.py
```diff
@@ -35,16 +37,48 @@ def validation_context(request, response=None):
...
         raise exception_response(206)
-        assert e.child is not None
     except Exception:
         raise exception_response(400)
 
 
 validation_ctx_path = 'tests.acceptance.request_test.validation_context'
 
 
+def test_echo_date_with_pyramid_swagger_renderer(test_app):
```
striglia: nice tests 👍

tests/acceptance/request_test.py
```diff
@@ -35,16 +37,48 @@ def validation_context(request, response=None):
...
+    exc = None
+    response = None
+    try:
+        response = test_app.post_json('/echo_date_json_renderer', input_object)
+    except TypeError as exception:
+        exc = exception
+
+    served_swagger_versions = test_app.app.registry.settings['pyramid_swagger.swagger_versions']
+
+    if '2.0' in served_swagger_versions:
```
striglia: Do we need one more test for 2.0 where we use the new renderer? The one above is testing 1.2 (at least according to the comment)

docs/quickstart.rst
macisamuele: I copied / got inspired by https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#adding-a-new-renderer .

pyramid_swagger/renderer.py
macisamuele: ``SwaggerMappingError`` could be raised if (the list is not exhaustive but gives an idea about the implications)
 - the response spec is not specifying the type (which is weird but possible, especially if no response validation is defined) -> https://github.com/Yelp/bravado-core/blob/619dec76ecbb67858129b808e8af4e5f8496e7a6/bravado_core/marshal.py#L60
 - the response body is not conform to the response spec (attribute ``attr`` is specified to be of type ``object``, but the value to marshal has ``attr`` set to ``1``)  -> https://github.com/Yelp/bravado-core/blob/619dec76ecbb67858129b808e8af4e5f8496e7a6/bravado_core/marshal.py#L133
 - An ``attr`` is defined  as type ``string`` and format ``date`` but the view returned object has ``attr`` set to date string representation and not ``datetime.date`` instance -> https://github.com/Yelp/bravado-core/blob/619dec76ecbb67858129b808e8af4e5f8496e7a6/bravado_core/formatter.py#L43

The idea of acting as NO OP is because this cases should be already handled by the downstream renderer or the service is returning a not spec compliant response. (which is the current behaviour)

We can eventually make sure that this conditions are made more visible maybe with WARNING logs ... what do you think?

tests/acceptance/request_test.py
macisamuele: ``test_app`` is generated by https://github.com/striglia/pyramid_swagger/blob/8cb3d30959006a85777374847b9aae8308dd98c3/tests/acceptance/request_test.py#L13-L15 and generates 3 tests with support for Swagger 1.2 only, Swagger 2.0 only and Swagger 1.2 and 2.0 

The idea is that in case Swagger 2.0 support is present we're going to receive an exception from ``JSON`` renderer because it is not able to properly render a ``datetime.date`` instance.
The case of ``Swagger1.2`` support only is different because in case of Swagger1.2 support only ``request.swagger_data`` is equivalent to ``request.json`` according to https://github.com/striglia/pyramid_swagger/blob/25fc16c9070bb39cec602b0374977e4bbd2c29a3/pyramid_swagger/tween.py#L337-L352
You can see that I'm asserting that the field is a string (https://github.com/macisamuele/pyramid_swagger/blob/maci-pyramid_swagger-renderer/tests/acceptance/app/__init__.py#L45)

pyramid_swagger/renderer.py
striglia: Hm fair point on not wanting to change current behavior. Maybe let's keep an eye out for whether this needs logging as we go forward? No need to add it now, considering we don't log anywhere else yet :<

tests/acceptance/request_test.py
striglia: Ah, the parameterizing makes sense then

macisamuele:
I have: 
 * rebased the branch on top of the latest master (sorry if seems that there are a lot of changes)
 * applied @sjaensch and @striglia feedbacks
 * updated changelog

macisamuele:
I have:
 * updated changelog
 * applied @sjaensch and @striglia feedbacks
 * rebased this branch on top of the most recent master (avoid merge conflicts with changelog)

striglia:
No objections here. Let's get it out there and see how it works in practice ⛵

PR #217, opened 2017-09-13
Adding support of new Bravado param `include_missing_properties`

python4g:
In some cases we need to use [JSON Merge Patch Format](https://tools.ietf.org/html/rfc7386) for PATCH resource. If we need to update only person's last name we send `{'last_name': 'Name'}` ignoring other fields but current logic set `None` to all fields that weren't included into request. 
Bravado added special parameter `include_missing_properties` to handle such cases so it would be great to add support of this config property into Your library.

Default value is `True` so I updated test to handle `False` case.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -55,6 +55,11 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         # Default: False
         pyramid_swagger.use_models = false
 
+        # Set value for parameter defined in swagger schema to None if value was not provided.
+        # Skip parameter if value is missed and include_missing_properties is False.
+        # Default: True
+        pyramid_swagger.include_missing_properties = true
+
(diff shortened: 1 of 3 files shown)


striglia:
Seems reasonable. Can I ask you to add an entry to the RST file for http://pyramid-swagger.readthedocs.io/en/latest/configuration.html to explain its purpose? If easier, feel free to just write a brief sentence and delegate the rest to any relevant bravado-core docs/issues.

macisamuele:
@striglia I was working on making documentation more understandable (even for who didn't wrote the code 😄 ) ... hopefully by the end of today I can publish something

striglia:
Ah evening Maci :) Improving docs here or on the bravado-core side? Either way 👍

python4g:
@striglia done :bowtie:

striglia:
Thanks for the contribution!

PR #218, opened 2017-09-24
Convert enable_api_doc_views to bool

DaineDanielson:
Previously, setting
pyramid_swagger.enable_api_doc_views = false
in a configuration .ini file would not disable the api_doc views,
because the string 'false' is truthy.

 - Convert the string-value of
   settings.get('pyramid_swagger.enable_api_doc_views', True)
   to a bool using pyramid.settings.asbool .
   (https://docs.pylonsproject.org/projects/pyramid/en/latest/api/settings.html)

### DIFF
--- a/pyramid_swagger/__init__.py
+++ b/pyramid_swagger/__init__.py
@@ -42,7 +42,8 @@ def includeme(config):
         under=pyramid.tweens.EXCVIEW
     )
 
-    if settings.get('pyramid_swagger.enable_api_doc_views', True):
+    if pyramid.settings.asbool(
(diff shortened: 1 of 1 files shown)


striglia:
Thanks for the fix Daine! This isn't the first bug we've had loading ini files....do you have any idea how we can effective test ini features? I'd love to write a test module for making sure we properly load from file...

DaineDanielson:
I don't have a thought of how to do this off the top of my head, but it would definitely be great to find a way to test this in general, for any ini-setting, existing or future. I would need to get a broader view of the pyramid_swagger codebase, which I sadly don't have time to do right now. Thanks for being so responsive on this, @striglia!

PR #221, opened 2017-10-23
bravado-core passthrough configs

### LINKED ISSUE #220: Improve how pyramid_swagger builds bravado-core configs
The current implementation of [``create_bravado_core_config``](https://github.com/striglia/pyramid_swagger/blob/c97c227f2a103b73fbf72906cac977df114ad622/pyramid_swagger/ingest.py#L188) limits the bravado-core configurations that could be set. The project already had some issues asking to expose new bravado-core configs through pyramid_swagger configs (ie. #211).

I'm wondering if we could actually have a passthrough prefix for setting bravado-core configs.
The idea is to use all the pyramid_swagger configs that starts with a given prefix as bravado-core configs like ``bravado_core.XXX: 1`` will set bravado-core config XXX to 1

macisamuele:
Fixes #220.

This PR deprecates (with a warning only for now) the usage of ``pyramid_swagger.xxx`` configs for the configuration of the underlying ``bravado_core`` instance.
In order to _trick_ bravado-core related configs ``bravado_core.xxx`` should be used instead.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -97,6 +97,12 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         pyramid_swagger.dereference_served_schema = false
 
 
+.. note::
+
+    ``pyramid_swawgger`` uses a ``bravado_core.spec.Spec`` instance for handling swagger related details.
+    You can set `bravado-core config values <http://bravado-core.readthedocs.io/en/stable/config.html>`_ by adding a ``bravado-core.`` prefix to them.
+
+
 Note that, equivalently, you can add these settings during webapp configuration:
 
 .. code-block:: python
--- a/pyramid_swagger/ingest.py
+++ b/pyramid_swagger/ingest.py
@@ -3,10 +3,12 @@ from __future__ import unicode_literals
 
 import glob
 import os.path
+import warnings
 
 import simplejson
 from bravado_core.spec import build_http_handlers
 from bravado_core.spec import Spec
+from six import iteritems
 from six.moves.urllib import parse as urlparse
 
 from .api import build_swagger_12_endpoints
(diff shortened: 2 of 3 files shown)


docs/configuration.rst
```diff
@@ -97,6 +97,12 @@ A few relevant settings for your `Pyramid .ini file <http://docs.pylonsproject.o
         pyramid_swagger.dereference_served_schema = false
 
 
+.. note::
+
+    ``pyramid_swawgger`` uses a ``bravado_core.spec.Spec`` instance for handling swagger related details.
+    It is possible to tune ``bravado-core`` configs adding ``bravado_core.`` prefix to the ``bravado-core`` config keys.
```
sjaensch: "You can set \`bravado-core config values \<http://bravado-core.readthedocs.io/en/stable/config.html>`_ by adding a \`\`bravado-core.\`\` prefix to them."

pyramid_swagger/ingest.py
```diff
@@ -16,6 +18,10 @@
 from .spec import validate_swagger_schema
 
 
+# Prefix of configs that will be passed to underline bravado-core instance
```
sjaensch: to the underlying bravado-core instance

tests/ingest_test.py
```diff
@@ -120,8 +121,17 @@ def test_create_bravado_core_config_with_defaults():
...
+def test_create_bravado_core_config_non_empty_deprecated_configs(mock_warnings):
     some_format = mock.Mock(spec=SwaggerFormat)
+    expected_bravado_core_config = {
+        'validate_requests': True,
+        'validate_responses': False,
+        'validate_swagger_spec': True,
+        'use_models': True,
+        'formats': [some_format],
+        'include_missing_properties': False
+    }
```
sjaensch: Should we move this part to a fixture so it can be shared between tests?

PR #226, opened 2018-03-11
swagger UI: add support for embedded Swagger UI API explorer

ergo:
Hi,

I've added support for embedded Swagger UI API explorer along with tests and basic documentation.


ppsimatikas:
When is this going to be merge?
I like the idea of embedded swagger UI.

Also, It would be great if the JSON error responses were enabled through a configuration option like:

`pyramid_swagger.error_reponse_format = json or html`

PR #229, opened 2018-05-17
Fix tests due to bravado core update

macisamuele:
Latest changes in bravado-core will not recognise objects that does not define `"type": "object"` as SCHEMA objects, so they won't be present into `#/definitions` for the flattened specs. As objects without type object are added inline, in case of recursive definition the flattening process will produce an unbounded recursion and so the test failure. 

This causes the two tests failure on master:
* `test_dereferenced_swagger_schema_bravado_client` : fixed by adding type object in the used specs
* `test_extenal_conflicting_namespaces` : has been removed as this checks internals of bravado-core


macisamuele:
@striglia , @sjaensch : could you have a look to this? It would be nice to merge this so travis could be green again.

PR #231, opened 2018-05-23
pyramid-swagger should not assume that body is present

macisamuele:
While working with endpoints which specify an optional body parameter pyramid-swagger raises a really ugly exception related to the fact that the library is trying to decode an empty body.
The raised exception looks like
```
self = <json.decoder.JSONDecoder object at 0x10e811198>, s = '', idx = 0

    def raw_decode(self, s, idx=0):
        """Decode a JSON document from ``s`` (a ``str`` beginning with
            a JSON document) and return a 2-tuple of the Python
            representation and the index in ``s`` where the document ended.

            This can be used to decode a JSON document from a string that may
            have extraneous data at the end.

            """
        try:
            obj, end = self.scan_once(s, idx)
        except StopIteration as err:
>           raise JSONDecodeError("Expecting value", s, err.value) from None
E           json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
NOTE: the stacktrace above is obtained by running `python -m pytest tests/acceptance/request_test.py::test_post_endpoint_with_optional_body[1.2-body0]` on sha `79a5cfa2c57b02dbbe351ffc5eaa48b157542f88`

In order to prevent unwanted JSON decoding we need to ensure that the `PyramidSwaggerRequest` will not call `self.request.json_body` if the body is emtpy.

While writing checking tests I've noticed that object validation fails either on Swagger 1.2 and Swagger  2.0.
Swagger1.2 validation is handled by this library and the fix is alraedy present on this PR (otherwise tests would have been red), while the Swagger2.0 validation is performed by bravado-core.

A PR will be opened for that project too, but in the mean time Swagger 2.0 allows us to disable request validation if needed.

### DIFF
--- a/pyramid_swagger/tween.py
+++ b/pyramid_swagger/tween.py
@@ -268,7 +268,10 @@ class PyramidSwaggerRequest(IncomingRequest):
         return result
 
     def json(self, **kwargs):
-        return getattr(self.request, 'json_body', {})
+        if self.request.is_body_readable:
+            return getattr(self.request, 'json_body', {})
+        else:
+            return None
 
 
 class PyramidSwaggerResponse(OutgoingResponse):
--- a/tests/acceptance/request_test.py
+++ b/tests/acceptance/request_test.py
@@ -1,18 +1,23 @@
 # -*- coding: utf-8 -*-
 import datetime
 from contextlib import contextmanager
+from json import dumps
 
 import pytest
 import simplejson
 from _pytest.fixtures import FixtureRequest
 from mock import Mock
 from pyramid.httpexceptions import exception_response
+from webtest.utils import NoDefault
 
 from pyramid_swagger import exceptions
 
 
 # Parameterize pyramid_swagger.swagger_versions
-@pytest.fixture(params=[['1.2'], ['2.0'], ['1.2', '2.0']])
+@pytest.fixture(
+    params=[['1.2'], ['2.0'], ['1.2', '2.0']],
+    ids=['1.2', '2.0', '1.2-2.0'],
+)
 def test_app(request, **overrides):
     """Fixture for setting up a test test_app with particular settings."""
     from .app import main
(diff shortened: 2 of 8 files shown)


pyramid_swagger/tween.py
```diff
@@ -268,7 +268,10 @@ def files(self):
         return result
 
     def json(self, **kwargs):
-        return getattr(self.request, 'json_body', {})
+        if self.request.is_body_readable:
+            return getattr(self.request, 'json_body', {})
+        else:
+            return None
```
sjaensch: Why are you using `None` here instead of the empty dictionary like above? It seems to me like `None` would be the better return value, but I think it should be consistent (and changing it might be backwards incompatible).

tests/acceptance/request_test.py
```diff
@@ -79,6 +84,13 @@ def test_echo_date_with_json_renderer(test_app):
         assert response.json == input_object
 
 
+@pytest.mark.parametrize('body', [NoDefault, {}])
+def test_post_endpoint_with_optional_body(test_app, body):
+    expected_length = len(dumps(body)) if body is not NoDefault else 0
```
sjaensch: Shouldn't this be a parameter as well?

tests/acceptance/request_test.py
```diff
@@ -79,6 +84,13 @@ def test_echo_date_with_json_renderer(test_app):
...
+    # This is brutal for now ... it will be used to at first for debugging the issue
```
sjaensch: I don't get this comment.

pyramid_swagger/tween.py
macisamuele: I think that defaulting the json_body to empty dictionary is not correct as the body of a request could potentially be of `type: string`.

About consistency the main concern is that we use `None` as sentinel for missing parameter.
If we return something that is not None that value will eventually be validated and it could cause validation error for a parameter that has not been defined in the request


Anyway that was used to handle the case that `self.request` does not have `json_body` attribute. `self.request` comes from `WebOb` library (which has `json_body` defined by really old versions), so there should be no instances on which line 272 returns the default value

tests/acceptance/request_test.py
macisamuele: ... testing leftover :)
The test was really brutal at first ... if the call does not raise error I'm happy :)

tests/acceptance/request_test.py
macisamuele: Doing it

PR #232, opened 2018-05-31
Fix tests failure related to x-model presence

macisamuele:
Due to the latest changes in bravado-core the specs used for testing will have x-model in the swagger file returned by pyramid-swagger.

I'm addressing this by ignoring the presence of x-model in the response.

### DIFF
--- a/tests/acceptance/yaml_test.py
+++ b/tests/acceptance/yaml_test.py
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
 import base64
+import copy
 import json
 
 import pytest
 import yaml
+from bravado_core.schema import is_dict_like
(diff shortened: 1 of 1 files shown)


sjaensch:
Shouldn't we change the `expected_dict` to contain the x-model markers? Why did you decide to modify the output instead?

macisamuele:
As we spoke on person we should avoid to "byte-check" all the details related to bravado-core; does not feel correct that bravado-core changes break pyramid-swagger tests.

As mentioned is important to show that pyramid swagger does not necessarily return the spec as on disk, but I think that tests in https://github.com/striglia/pyramid_swagger/blob/master/tests/acceptance/relative_ref_test.py will be enough to highlight that

PR #234, opened 2018-06-30
Add appveyor support and fix issues on Windows 

### LINKED ISSUE #233: Cannot open swagger.json due to invalid path
In ingest.py, in line with code
`schema_url = urlparse.urljoin('file:', os.path.abspath(schema_path))`
the result schema_url in my case was `c:/path/to/file/swagger.json` instead of `file:c:/path/to/file/swagger.json`, which was resulting in error later when trying to open this file with `urlopen`

As a temporary fix, I am currently concating strings instead of using `urlparse.urljoin`

macisamuele:
fixes #233

This PR adds support to run tests for Windows platform on appveyor and fixes the issues related to paths and URIs handling on windows.

Using `pathname2url` while transforming a path to an URI ensures that on windows paths are correctly formatted while on linux and mac the function looks like a no-op.

There are 2 tests modifications:
 * `tests/acceptance/relative_ref_test.py::test_dereferenced_swagger_schema_retrieval`: this is needed as bravado-core flattening produces a different swagger spec dict in case it runs on windows platform
* `tests/spec_test.py::test_proper_error_on_missing_api_declaration`: this is needed as `\` in the exception string is escaped and it's harder to regenerate the same string and keeping the test easy to read. The approach used was to check that directory name and file name are present on the exceptio

Tests pass on Windows: https://ci.appveyor.com/project/macisamuele/pyramid-swagger/build/1

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,28 @@
+version: '{build}'
+image: Visual Studio 2017
+
+environment:
+  matrix:
+  # Available python versions and their locations on https://www.appveyor.com/docs/build-environment/#python
+  - PYTHON: C:\Python27-x64
+    TOXENV: py27
+  - PYTHON: C:\Python27-x64
+    TOXENV: py27-pyramid15
+  - PYTHON: C:\Python35-x64
+    TOXENV: py35
+  - PYTHON: C:\Python36-x64
+    TOXENV: py36
+
+build: off
+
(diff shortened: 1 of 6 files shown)


striglia:
Added the badge to the README and created https://ci.appveyor.com/project/striglia/pyramid-swagger for regular builds

PR #237, opened 2019-05-05
Fix tests and make pyramid-swagger compatible with jsonschema>3

macisamuele:
The goal of this CR is to fix tests that are red from ever (roughly 6 months).
The failures were caused by:
 * pre-commit updates (flake8 version was not pinned)
 * release of jsonschema>3 (the library was using private internals which have changed)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,3 +27,6 @@ repos:
     rev: v1.4.0
     hooks:
     -   id: reorder-python-imports
+        args:
+        - --add-import
+        - from __future__ import absolute_import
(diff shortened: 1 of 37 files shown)


sjaensch:
Tests are not passing on Windows, wanna fix that first?

macisamuele:
@sjaensch the test failures on AppVeyor (windows) are caused by bravado-core.
My suggestion here would be to accept those changes for now, in the mean time I've prepared a CR to deal with the issue

striglia:
Many thanks Samuele -- I clearly let this one lapse 😅

PR #243, opened 2019-08-26
Fix exception assertions' content

macisamuele:
The goal of this PR is to fix the failing tests.


macisamuele:
@sjaensch For what I see travis executed daily, while appveyor is executed only in case of changes.
Checking the first time pyramid-swagger failed to run tests on master ([travis](https://travis-ci.org/striglia/pyramid_swagger/jobs/552261968)) I see that they started to fail once pytest>=5 was released

PR #245, opened 2020-06-14
Convert bool values from a .ini file

hiromu:
When I use a .ini file for configuration, some variables are passed to bravado-core in the form of str and always treated as `True`.
Since all variables except `formats` are boolean, they should be converted using `asbool()`: https://bravado-core.readthedocs.io/en/stable/config.html

The same thing happens for `enable_api_doc_views` in `__init__.py`, and thus I fixed.

### DIFF
--- a/pyramid_swagger/ingest.py
+++ b/pyramid_swagger/ingest.py
@@ -8,6 +8,7 @@ import os.path
 import simplejson
 from bravado_core.spec import build_http_handlers
 from bravado_core.spec import Spec
+from pyramid.settings import asbool
 from six import iteritems
 from six.moves.urllib import parse as urlparse
 from six.moves.urllib.request import pathname2url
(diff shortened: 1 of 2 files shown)


pyramid_swagger/ingest.py
```diff
@@ -217,12 +218,13 @@ def create_bravado_core_config(settings):
         'use_models': False,
     }
     configs.update({
-        bravado_core_key: settings[pyramid_swagger_key]
+        bravado_core_key: (settings[pyramid_swagger_key] if bravado_core_key == 'formats'
+                           else asbool(settings[pyramid_swagger_key]))
```
sjaensch, line 222: This is dangerous, as any new bravado-core config key which doesn't have a boolean value would break this. We should instead list the bravado-core keys we know of and the function we should use for each key to get the value. Ideally we'd provide an API to do this within bravado-core.

pyramid_swagger/ingest.py
```diff
@@ -217,12 +218,13 @@ def create_bravado_core_config(settings):
...
         for pyramid_swagger_key, bravado_core_key in iteritems(config_keys)
         if pyramid_swagger_key in settings
     })
     configs.update({
-        key.replace(BRAVADO_CORE_CONFIG_PREFIX, ''): value
+        key.replace(BRAVADO_CORE_CONFIG_PREFIX, ''): asbool(value)
```
sjaensch, line 227: This doesn't handle `formats`.

hiromu:
I understand your concerns on the forward compatibility.

Are you considering to implement an API within bravado-core? Otherwise, I'll update this pull request so as to use a hard-coded allowlist to determine whether to apply `asbool`. 

> how were you specifying config values that you ran into this issue?

In my implementation, I'm specifying `bravado_core.internally_dereference_refs = false` in environment.ini and just calling `config = Configurator(settings=settings); config.include('pyramid_swagger')` in `main(global_config, **settings)` of Pyramid.

PR #248, opened 2024-04-26
Remove swagger 1.2 support

raychan:
When running tests on python 3.8 and 3.10, tests are failing for Swagger 1.2
Since we are mostly on swagger 2.0, I think we can stop supporting swagger 1.2


danielpops (approved):

tests/acceptance/response_test.py
```diff
@@ -61,37 +66,45 @@ def get_registry(settings):
...
+    op = spec.get_op_for_request(request.method, path_pattern)
+    mock_route_info = {'match': request.matchdict, 'route': None}
+    mock_route_mapper = Mock(spec=IRoutesMapper, return_value=mock_route_info)
+    with patch('pyramid_swagger.tween.get_op_for_request', return_value=op):
+        with patch('pyramid.registry.Registry.queryUtility',
+                   return_value=mock_route_mapper):
+            validation_tween_factory(handler, registry)(request)
 
 
 def test_response_validation_enabled_by_default():
```
jtwang, line 110: Not your problem since this code is preexisting, but just for the record, I'm not a big fan of this test existing on its own. It should be a part of a series that tests validation passes with a valid response, then fails, then test enabled vs disabled. Also there should be wrappers around the test requests.

Anyways.... _gets off soapbox_

tests/acceptance/response_test.py
```diff
@@ -163,13 +191,17 @@ def test_500_when_response_arg_is_wrong_type():
     )
     response = Response(
         body=simplejson.dumps({
-            'raw_response': 1.0,
+            'raw_response': 1.0,  # <-- is supposed to be a string
```
jtwang, line 194: comment++

tests/acceptance/response_test.py
```diff
@@ -197,29 +234,19 @@ def test_200_for_good_validated_array_response():
...
-        swagger_versions=['1.2'],
+    assert build_test_app(
+        swagger_versions=['2.0'],
         **{'pyramid_swagger.enable_response_validation': True}
-    )
-    response = app.post_json('/sample', {'foo': 'test', 'bar': 'test'})
-    assert response.status_code == 200
-
-
-def test_200_skip_validation_for_excluded_path():
```
jtwang, line 212: deletion++

tests/api_test.py
```diff
@@ -3,60 +3,10 @@
...
-
-def build_config(schema_dir):
-    return mock.Mock(
-        registry=get_registry({
-            'swagger_schema': get_swagger_schema(schema_dir),
-        })
-    )
-
-
-def test_proper_error_on_missing_resource_listing():
```
jtwang, line 39: Do we want to replace this with a test that checks the exception thrown if `swagger.json` is missing?

tests/tween_test.py
```diff
@@ -119,46 +102,6 @@ def test_validation_skips_path_properly():
     assert not should_exclude_path(excluded_paths, '/sample/resources')
 
 
-# TODO: Should probably be migrated to acceptance tests after we make mocking
-# schemas easier there.
-def test_validation_content_type_with_json():
```
jtwang, line 124: I'm a little lost - did these already get migrated and so OK to delete?

jtwang (commented):
Thanks for tackling this! It's a much bigger change than I'd expected. :S I have some questions for you, see below. :)

tests/acceptance/response_test.py
```diff
@@ -61,37 +66,45 @@ def get_registry(settings):
...
+
     registry = get_registry(settings)
 
-    # Let's make request validation a no-op so we can focus our tests.
-    with mock.patch.object(pyramid_swagger.tween, 'validate_request'):
-        validation_tween_factory(handler, registry)(request)
+    # This is a little messy because the current flow of execution doesn't
+    # set up the route_info in pyramid. Have to mock out the `route_info`
+    # so that usages in the tween meet expectations. Holler if you know a
+    # better way to do this!
```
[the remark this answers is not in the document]
raychan: Original there were two test files, one is response_test.py that is for swagger 1.2 and one is response20_test.py that is for swagger 2.0.
I deleted response_test.py and renamed response20_test.py to response_test.py, but if this is hard to review, I can undo the rename

tests/tween_test.py
raychan: If I understand it correctly, `tween.validate_response` was only used by swagger 1.2 and is deleted in this PR here:
https://github.com/Yelp/pyramid_swagger/pull/248/files#diff-12cf5562318320d89b816f8122b3b384ede6cae269a0a5760d216ce6494465c9L544

jtwang (approved):

raychan:
Close in favor of https://github.com/Yelp/pyramid_swagger/pull/249

PR #253, opened 2024-05-06
Add response_validation_exclude_routes support

raychan:
We want to add the ability to exclude routes for only response validation (while keep request response)

### DIFF
--- a/pyramid_swagger/__about__.py
+++ b/pyramid_swagger/__about__.py
@@ -13,7 +13,7 @@ __title__ = "pyramid_swagger"
 __summary__ = "Swagger tools for use in pyramid webapps"
 __uri__ = "https://github.com/striglia/pyramid_swagger"
 
-__version__ = "2.9.0"
+__version__ = "2.8.0"
 
 __author__ = "Scott Triglia"
 __email__ = "scott.triglia@gmail.com"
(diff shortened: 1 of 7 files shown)


pyramid_swagger/tween.py
```diff
@@ -462,6 +468,11 @@ def should_exclude_request(settings, request, route_info):
     )
 
 
+def should_exclude_response_validation(settings, route_info):
+    print(route_info)
```
hippoandy, line 472: Remove this.

hippoandy (approved):
LGTM!

pyramid_swagger/__about__.py
```diff
@@ -13,7 +13,7 @@
 __summary__ = "Swagger tools for use in pyramid webapps"
 __uri__ = "https://github.com/striglia/pyramid_swagger"
 
-__version__ = "2.8.0"
```
ymilki, line 16: Can you seperate the version change and changelog change into follow-up PR? This way, the changes can be treated independently of the version change.

ymilki (changes requested):

pyramid_swagger/__about__.py
raychan: done

hippoandy (approved):

ymilki (approved):

